PR #12, opened 2016-12-07
Add variables to specify which project(s) to restore and run. Start app using assembly.

tmds:
DOTNET_RESTORE_ROOT can be used to specify what project(s) to restore.
DOTNET_STARTUP_PROJECT is used to specify what project to run.

The application is started by invoking `dotnet <assembly>` instead of `dotnet run`.
`dotnet run` introduces a process-in-the-middle which causes side effects like
buffering stdin&stdout and intercepting signals. By eliminating this intermediary
process, the app will run as PID 1.


jerboaa:
I'll have a look at this later today.

jerboaa:
We should have those items covered in run/test. In particular the DOTNET_RESTORE_ROOT which takes a "list" should get tested with both. A single project and a list of projects. We probably have to come up with some contrived example apps to do this.

jerboaa:
Thanks for adding the tests. It worked fine here. Please squash the commits and I'll merge.

tmds:
GitHub can do the squashing.

jerboaa:
Thanks, merged.

jerboaa:
[skipci]

PR #16, opened 2016-12-09
Add tests for CLI apps and top-level build.sh

jerboaa:
Also add top-level build.sh so as to be able to drive
building/testing of all versions.

Example building all images, running tests on them:

  $ sudo ./build.sh

Example building 1.0 image and running tests on it:

  $ sudo VERSIONS=1.0 ./build.sh

PR #17, opened 2016-12-09
Add test for DOTNET_FRAMEWORK env var.

jerboaa:
The idea is to have one failing test if `-f <framework>` is not being passed to `dotnet build` in `s2i/assemble`. It fails on the s2i build, prior #11 and passes after.

This patch depends on #16.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+project.lock.json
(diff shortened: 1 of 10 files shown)


tmds:
LGTM

jerboaa:
[skipci]

PR #24, opened 2016-12-21
Adjust test for 1.0.3 update.

jerboaa:
1.0.3 has been released since, which reports itself as
1.0.0-preview2-003156.


jerboaa:
skipci

PR #27, opened 2016-12-22
Add dotnet-pgsql-persistent template

tmds:


jerboaa:
How would I test this? Could you please provide some documentation as to how this is intended to be used?

jerboaa:
[skipci]

tmds:
> How would I test this?

No idea. Do you know how the other templates are tested?

> Could you please provide some documentation as to how this is intended to be used?

Templates are supposed to be self-documenting. So give it a try, and you perhaps you find some places where the documentation can improve.

jerboaa:
Please don't assume users will know how to use it. I disagree on the self-documenting note. I'm sure you've run some commands/steps to test this. Documenting them here for others to reproduce would be nice.

What other templates are you referring to? We currently only have an imagestream file.

tmds:
> Please don't assume users will know how to use it. I disagree on the self-documenting note. I'm sure you've run some commands/steps to test this. Documenting them here for others to reproduce would be nice.
> 
> What other templates are you referring to? We currently only have an imagestream file.

This is an OpenShift template (https://docs.openshift.org/latest/dev_guide/templates.html).

It is similar to other templates OpenShift provides out of the box: cakephp-mysql-persistent, rails-pgsql-persistent, dancer-mgsql-persistent, ...

jerboaa:
General comment:

Please consider this from the point of a reviewer. For example, I as a reviewer I first look at the title and read the description next. My experience here was:

`Title: Add dotnet-pgsql-persistent template`
`Description: <none>`

At that point I'm already not super eager to look any further. We should aim at making work easier for reviewers, not harder. The suggestions for future PRs is:

- Make an effort to produce good summaries and descriptions. Examples:
  https://lkml.org/lkml/2016/12/22/310
  https://github.com/openshift/origin/pull/12324
- Add a note how this can be tested and/or what you've done to test it.

tmds:
@jerboaa my bad, I thought the title would be enough for you to figure out what it was about.

I tried this with the published dotnet:1.1 and a branch that works around the various issues we have fixed in the meanwhile.

tmds:
@jerboaa can you review this?

jerboaa:
@tmds I can review it, but not sure when. If somebody else could have a look meanwhile it would be helpful.

templates/dotnet-pgsql-persistent.json
```diff
@@ -0,0 +1,479 @@
...
+                        "ref": "${SOURCE_REPOSITORY_REF}"
+                    },
+                    "contextDir": "${CONTEXT_DIR}"
+                },
+                "strategy": {
+                    "type": "Source",
+                    "sourceStrategy": {
+                        "from": {
+                            "kind": "ImageStreamTag",
+                            "namespace": "${NAMESPACE}",
```
jerboaa: We should remove this namespace since this prevents users from using this template if the dotnet images come from an imagestream in the local (current) namespace and don't have it in `openshift`.

tmds: Is this different for the rails example? I got this from their template: https://github.com/openshift/rails-ex/blob/master/openshift/templates/rails-postgresql-persistent.json#L102

jerboaa: They assume it's preloaded, I suppose. Which they probably do for all sclorg images. Rails is one of them. We should not (yet) assume that. It should work either way if we omit the namespace.

tmds:
`DOTNET_RESTORE_ROOT` must be added to the template to reduce the restore time.

jerboaa:
[skipci]

templates/dotnet-pgsql-persistent.json
```diff
@@ -0,0 +1,478 @@
...
+                                    {
+                                        "name": "Data__DefaultConnection__ConnectionString",
+                                        "value": "Host=${DATABASE_SERVICE_NAME};Database=${DATABASE_NAME};Username=${DATABASE_USER};Password=${DATABASE_PASSWORD}"
+                                    }
+                                ],
+                                "resources": {
+                                    "limits": {
+                                        "memory": "${MEMORY_LIMIT}"
+                                    }
+                                }
```
jerboaa: Could you please add a readiness/liveness probe like this to the template? This avoids a message in the webconsole showing up that the deployment does not have one.

    
                                "livenessProbe": {
                                    "httpGet": {
                                        "path": "/Store",
                                        "port": 8080,
                                        "scheme": "HTTP"
                                    },
                                    "initialDelaySeconds": 40,
                                    "timeoutSeconds": 10,
                                    "periodSeconds": 10,
                                    "successThreshold": 1,
                                    "failureThreshold": 3
                                },
                                "readinessProbe": {
                                    "httpGet": {
                                        "path": "/Store",
                                        "port": 8080,
                                        "scheme": "HTTP"
                                    },
                                    "initialDelaySeconds": 10,
                                    "timeoutSeconds": 30,
                                    "periodSeconds": 10,
                                    "successThreshold": 1,
                                    "failureThreshold": 3
                                },

jerboaa:
This looks OK to me, fwiw. @tmds feel free to merge.

tmds:
@jerboaa Let's merge #60 first. I'll handle the merge conflict on this PR and update the README to describe this template too.

jerboaa:
Do you mind adding some sample usage of the template here and reference in README.md (of this repo) like I did with #64:
https://github.com/redhat-developer/s2i-aspnet-musicstore-ex/blob/rel/1.1.0-preview1/README.md

Note there are also references to gpiercey which should be replaced by redhat-developer. Thanks!

README.md
```diff
@@ -49,12 +49,27 @@ Example: `dotnet/dotnetcore-10-rhel7`
...
 oc create -f <template.json>
 ```
 
+To instantiate a template you can use the `oc new-app` command:
+
+```
+oc new-app --template=<template>
+```
+
+In case you imported the template in your project, set the `NAMESPACE` parameter to your project name
```
jerboaa: Not sure if that's correct. `NAMESPACE` needs to be specified when the dotnet imagestreams are *not* in `openshift`, but in a local project.

Suggested wording:

In case you've imported the [.NET Core imagestreams](https://github.com/redhat-developer/s2i-dotnetcore/blob/master/dotnet_imagestreams.json) into your local project, `NAMESPACE` needs to be specified to match it. This can be done by setting the `NAMESPACE` parameter via `-p NAMESPACE=<project>`.

README.md
```diff
@@ -49,12 +49,27 @@ Example: `dotnet/dotnetcore-10-rhel7`
...
+by adding the `-p NAMESPACE=<project>` parameter.
+
+The template can also be instantiated using the OpenShift web console. Login to the console and
+navigate to the desired project. Click the **Add to Project** button. Search and select the \<template>.
```
jerboaa: I'm not sure whether `\<template>` is very clear. How about this instead?

"Search and select the desired template by it's name, for example `dotnet-example`."

README.md
tmds: Right. I will remove the paragraph since the README isn't describing how to import image streams.

tmds:
@jerboaa can you do a test? When the app starts it should say `Using data provider: npgsql` which means it is connecting to PostgreSQL.
I did a test before I renamed the branch and added the print statement.

jerboaa:
@tmds working on it.

jerboaa:
@tmds Works fine for me. Ship it!

tmds:
Thanks @jerboaa

PR #30, opened 2017-01-05
Publish the dotnet application and run it from the published folder

tmds:
The publish command is intended to pack the application and all its dependencies in a folder. The user can perform some additional steps as part of the publish operation and specify the artifacts that need to be copied.
The published folder is then used as the working directory when executing the application. This is important because many applications set the ASP.NET ContentRoot to the current directory.


tmds:
The test is failing with:
```
dotnet <assembly>' not process with PID 1. PID 1 process was: '/bin/bash /usr/libexec/s2i/run'
```

Probably this is a race condition between the container startup and the pid check. dotnet may not be running yet.

Issue #28, opened 2016-12-22
$PWD and $HOME should be different


jerboaa:
For s2i container builds the current working directory coincides with $HOME of the container user (default). This is a problem because `dotnet build` recursively looks for source code files. If the app we are currently building pulls in nuget packages which contain source code files, `dotnet build` will try to compile them which might fail.

We need to make $HOME outside $PWD so as to not fall into the trap of trying to compile sources in downloaded nuget packages.

See also:
https://bugzilla.redhat.com/show_bug.cgi?id=1408146

PR #31, opened 2017-01-12
PWD should not be HOME.

jerboaa:
When the current working directory is the same as HOME this might
cause problems when building .NET Core apps. The problem stems
from some nuget packages including C# source files. When
"dotnet build" is invoked in a directory which is also the HOME
folder, it will attempt to recursively build all source files
under it. That might fail for some source files in $HOME/.nuget.

See issue #28 for details. The fix is to set HOME to one level
up. That is the current working directory will not be outside the
recursive search path when "dotnet build" is being invoked.

Closes issue #28.

### DIFF
--- a/1.0/Dockerfile.rhel7
+++ b/1.0/Dockerfile.rhel7
@@ -7,7 +7,7 @@ EXPOSE 8080
 ENV DOTNET_CORE_VERSION=1.0
 # Default to UTF-8 file.encoding
 ENV LANG=C.UTF-8 \
-    HOME=/opt/app-root/src \
+    HOME=/opt/app-root \
     PATH=/opt/app-root/src/.local/bin:/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
     STI_SCRIPTS_PATH=/usr/libexec/s2i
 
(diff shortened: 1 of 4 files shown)


jerboaa:
@tmds @omajid @gpiercey Could I get a review, please? Thanks!

jerboaa:
Thanks. I'll merge.

1.0/Dockerfile.rhel7
```diff
@@ -64,7 +64,7 @@ RUN cd /opt/app-root/src && mkdir cache-warmup && \
     cd .. && \
     rm -rf cache-warmup && \
     rm -rf /tmp/NuGetScratch && \
-    find .nuget -name '*.cs' -exec rm '{}' \;
+    find ${HOME}/.nuget -name '*.cs' -exec rm '{}' \;
```
tmds: Looks like this was a workaround solved by this PR. This line can be removed.

1.1/Dockerfile.rhel7
```diff
@@ -64,7 +64,7 @@ RUN cd /opt/app-root/src && mkdir cache-warmup && \
     cd .. && \
     rm -rf cache-warmup && \
     rm -rf /tmp/NuGetScratch && \
-    find .nuget -name '*.cs' -exec rm '{}' \;
+    find ${HOME}/.nuget -name '*.cs' -exec rm '{}' \;
```
tmds: idem

1.0/Dockerfile.rhel7
jerboaa: See #32 which removes this. Thanks!

PR #32, opened 2017-01-13
Remove unneeded work-around.

jerboaa:
After issue #31 the PWD does no longer
coincide with $HOME. Thus, a compile of
app sources no longer takes sources in
$HOME/.nuget into account. The workaround
for this can be removed.


jerboaa:
Thanks, merging.

PR #33, opened 2017-01-13
Add DEBUG variable to test/run.

jerboaa:
This allows one to run tests as follows:

$ sudo IMAGE_NAME=foo DEBUG=true ./test/run

By setting DEBUG to something non-empty bash
will print every line it executes. This is
useful for debugging tests.

$ sudo IMAGE_NAME=foo ./test/run

Would run tests as before (no debug output).

### DIFF
--- a/1.0/test/run
+++ b/1.0/test/run
@@ -1,13 +1,20 @@
 #!/bin/bash
 #
-# The 'run' performs a simple test that verifies that S2I image.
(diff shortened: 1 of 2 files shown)


jerboaa:
@tmds @omajid  anyone? This shouldn't be too controversial :)

tmds:
LGTM

PR #38, opened 2017-01-29
Fix build failure due to .NET framework target

tmds:
The hw_framework_config test includes a .NET framework target. When a dependency is added to the test project, the publish will fail with this message:

error DOTNET1012: The reference assemblies directory was not specified. You can set the location using the DOTNET_REFERENCE_ASSEMBLIES_PATH environment variable.

### DIFF
--- a/1.0/s2i/bin/assemble
+++ b/1.0/s2i/bin/assemble
@@ -16,6 +16,12 @@ cp -Rf /tmp/src/. ./
 echo "---> Installing dependencies ..."
 dotnet restore $DOTNET_RESTORE_ROOT
 
+# The publish operation fails when .NET framework targets are included in project.json
+# See https://github.com/dotnet/cli/issues/3636
+# This bug won't be fixed for project.json, workaround:
+export DOTNET_REFERENCE_ASSEMBLIES_PATH=/tmp/reference_assemblies
+mkdir -p $DOTNET_REFERENCE_ASSEMBLIES_PATH/.NETFramework/{v3.5,v4.0,v4.5,v4.5.1,v4.5.2,v4.6,v4.6.1,v4.6.2}
+
 echo "---> Building application from source ..."
 dotnet publish -f "$DOTNET_FRAMEWORK" -c "$DOTNET_CONFIGURATION" "$DOTNET_STARTUP_PROJECT"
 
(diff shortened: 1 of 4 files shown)


tmds:
Build log by adding dependency to hw_framework_config:
```
...
log  : Writing lock file to disk. Path: /opt/app-root/src/project.lock.json
log  : ./project.json
log  : Restore completed in 21670ms.
---> Building application from source ...
E0129 16:50:07.764020 04097 util.go:91] The current project is not valid because of the following errors:
E0129 16:50:07.764061 04097 util.go:91] /opt/app-root/src/project.json(10,15): error DOTNET1012: The reference assemblies directory was not specified. You can set the location using the DOTNET_REFERENCE_ASSEMBLIES_PATH environment variable.
E0129 16:50:07.764072 04097 util.go:91] 
E0129 16:50:08.830743 04097 main.go:352] An error occurred: non-zero (13) exit code from dotnet/dotnetcore-11-rhel7

```

1.0/s2i/bin/assemble
```diff
@@ -16,6 +16,12 @@ cp -Rf /tmp/src/. ./
 echo "---> Installing dependencies ..."
 dotnet restore $DOTNET_RESTORE_ROOT
 
+# The publish operation fails when .NET framework targets are included in project.json
+# See https://github.com/dotnet/cli/issues/3636
+# This bug won't be fixed for project.json, workaround:
+export DOTNET_REFERENCE_ASSEMBLIES_PATH=/tmp/reference_assemblies
+mkdir -p $DOTNET_REFERENCE_ASSEMBLIES_PATH/.NETFramework/{v1.1,v2.0,v3.5,v4.0,v4.0.3,v4.5,v4.5.1,v4.5.2,v4.6,v4.6.1,v4.6.2}
```
jerboaa: Is this being used for the `dotnet publish` step only? If so, could we remove the directories again after it? Otherwise all container instantiations of the image will have those.

tmds: is this also the case for '/tmp'?

jerboaa: Yes. This is one of the weird things to pay particular attention to wrt containers.

tmds: funky

tmds: Will s2i delete the /tmp/src folder?

jerboaa: On closer look we should check if this use of tmp during assemble actually results in files being present too in the resulting app-container. I'm pretty sure they persist if used (or created implicitly) in a Dockerfile instruction. What I'm trying to avoid is to have this cause some strange side-effect later down the line should it persist.

As to /tmp/src, I think it deletes it. Not sure.

tmds: @jerboaa the /tmp folder is persisted and the /tmp/src folder is not removed

jerboaa: Thanks for checking.

jerboaa:
Looks good. Merging.

PR #39, opened 2017-01-30
Make all DOTNET_ envvars, build envvars only

tmds:
Environment variables used by the assemble and run script, need to
be provided in the OpenShift build environment and deployment
environment.

We only uses the DOTNET_ environment variables in the
assemble script (build environment). The output is published
at /opt/app-root/publish including an s2i_run script to start
the .NET Core application.

### DIFF
--- a/1.1/Dockerfile.rhel7
+++ b/1.1/Dockerfile.rhel7
@@ -44,7 +44,7 @@ RUN yum install -y yum-utils && \
     yum install -y --setopt=tsflags=nodocs rh-dotnetcore11 nss_wrapper tar rh-nodejs4-npm && \
     yum clean all && \
     mkdir -p /opt/app-root/src /opt/app-root/publish && \
-    useradd -u 1001 -r -g 0 -d /opt/app-root -s /sbin/nologin \
+    useradd -u 1001 -r -g 0 -d /opt/app-root/src -s /sbin/nologin \
       -c "Default Application User" default && \
     chown -R 1001:0 /opt/app-root
 
--- a/1.1/Dockerfile.rhel7
+++ b/1.1/Dockerfile.rhel7
@@ -9,7 +9,9 @@ ENV DOTNET_CORE_VERSION=1.1
 ENV LANG=C.UTF-8 \
     HOME=/opt/app-root \
     PATH=/opt/app-root/src/.local/bin:/opt/app-root/src/bin:/opt/app-root/bin:/opt/app-root/node_modules/.bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
-    STI_SCRIPTS_PATH=/usr/libexec/s2i
+    STI_SCRIPTS_PATH=/usr/libexec/s2i \
+    DOTNET_PUBLISH_PATH=/opt/app-root/publish \
+    DOTNET_RUN_SCRIPT=/opt/app-root/publish/s2i_run
 
 LABEL io.k8s.description="Platform for building and running .NET Core 1.1 applications" \
       io.k8s.display-name=".NET Core 1.1" \
@@ -41,8 +43,8 @@ RUN yum install -y yum-utils && \
     yum-config-manager --enable rhel-7-server-ose-3.0-rpms && \
     yum install -y --setopt=tsflags=nodocs rh-dotnetcore11 nss_wrapper tar rh-nodejs4-npm && \
     yum clean all && \
-    mkdir -p /opt/app-root/src && \
-    useradd -u 1001 -r -g 0 -d /opt/app-root/src -s /sbin/nologin \
+    mkdir -p /opt/app-root/src /opt/app-root/publish && \
+    useradd -u 1001 -r -g 0 -d /opt/app-root -s /sbin/nologin \
       -c "Default Application User" default && \
     chown -R 1001:0 /opt/app-root
 
--- a/1.1/s2i/bin/run
+++ b/1.1/s2i/bin/run
@@ -3,21 +3,14 @@ source /opt/app-root/etc/generate_container_user
 
 set -e
 
-# User settable environment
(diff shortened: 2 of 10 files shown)


tmds:
@jerboaa @omajid ptal

1.1/s2i/bin/run
```diff
@@ -3,21 +3,14 @@ source /opt/app-root/etc/generate_container_user
...
-DOTNET_CONFIGURATION="${DOTNET_CONFIGURATION:-Release}"
-DOTNET_STARTUP_PROJECT="${DOTNET_STARTUP_PROJECT:-.}"
-
-# Private environment
-DOTNET_FRAMEWORK="netcoreapp1.1"
-APP_DLL_NAME="$(basename "$(realpath "${DOTNET_STARTUP_PROJECT}")").dll"
-APP_DLL_DIR="${DOTNET_STARTUP_PROJECT}/bin/${DOTNET_CONFIGURATION}/${DOTNET_FRAMEWORK}/publish"
+cd $DOTNET_PUBLISH_PATH
 
 ARG_FILE="arguments.txt"
```
tmds: @jerboaa do you know where 'arguments.txt' comes from? This is not a .NET pattern.
It isn't in the documentation. May it be removed?

1.1/Dockerfile.rhel7
```diff
@@ -41,8 +43,8 @@ RUN yum install -y yum-utils && \
     yum-config-manager --enable rhel-7-server-ose-3.0-rpms && \
     yum install -y --setopt=tsflags=nodocs rh-dotnetcore11 nss_wrapper tar rh-nodejs4-npm && \
     yum clean all && \
-    mkdir -p /opt/app-root/src && \
-    useradd -u 1001 -r -g 0 -d /opt/app-root/src -s /sbin/nologin \
+    mkdir -p /opt/app-root/src /opt/app-root/publish && \
+    useradd -u 1001 -r -g 0 -d /opt/app-root -s /sbin/nologin \
       -c "Default Application User" default && \
```
tmds: small fix: changed the -d argument to match with HOME

1.1/test/dotnetbot/project.json
```diff
@@ -13,5 +13,10 @@
     "netcoreapp1.1": {
       "imports": "dnxcore50"
     }
+  },
+  "publishOptions": {
+    "include": [
+      "arguments.txt"
+    ]
```
tmds: arguments.txt now needs to be a published file of the DOTNET_STARTUP_PROJECT
before it was a file in context-dir
To keep the old semantics, we'd need to treat it specially in the assemble script.

1.1/Dockerfile.rhel7
jerboaa: I'd prefer if we kept the user's home directory to be in the src folder. The reason for this is that when you get a terminal in the app container, the sources of the app will be in the current working directory. If we really want to change this, please do so in a separate PR, not lumping it in here. Thanks.

1.1/s2i/bin/run
jerboaa: It's just for convenience if there are many arguments to pass to an app. Pattern taken from some other container, that escapes me. It's been there from the start. If we are going to remove it we should perhaps give users some migration path.

jerboaa:
Can you elaborate the motivation about these changes? Is it to give users a way to debug problems easier?

tmds:
> Can you elaborate the motivation about these changes? Is it to give users a way to debug problems easier?

This makes it impossible to break the deployment by specifying DOTNET_* envvars.
Since the publish folder is outside of /src, the change makes it possible to remove the src folder.
It is also a step towards a runtime image.

1.1/s2i/bin/run
tmds: .NET has support for reading configuration in 'Microsoft.Extensions.Configuration'. There are providers that read Xml, Ini and Json configuration files. These are good alternatives for arguments.txt.
Remove in separate PR?

1.1/Dockerfile.rhel7
tmds: As it is intentional, I will revert the change.

1.1/test/dotnetbot/project.json
tmds: This change can be reverted when we drop support for arguments.txt.

1.1/test/qotd/project.json
```diff
@@ -16,7 +16,8 @@
   },
   "publishOptions": {
     "include": [
-      "quotes.txt"
+      "quotes.txt",
+      "arguments.txt"
     ]
```
tmds: This change can be reverted when we drop support for arguments.txt.

1.1/test/dotnetbot/project.json
tmds: This change can be reverted when we drop support for arguments.txt.

1.1/s2i/bin/run
tmds: > Remove in separate PR?

Thinking about it -> separate PR is preferable since it will also require removal of files and changing tests.

jerboaa: +1

1.1/Dockerfile.rhel7
jerboaa: I don't see this reverted, yet :(

jerboaa:
> This makes it impossible to break the deployment by specifying DOTNET_* envvars.
Since the publish folder is outside of /src, the change makes it possible to remove the src folder.
It is also a step towards a runtime image.

OK, this makes sense. I had to read the original commit message twice to fully grok this, though. Basically the environment variables will still be there, but they won't be used as the `run` script delegates to `s2i_run` which does not need/use them. Can I ask you to give such rationale right when posting the PR?

tmds:
> Can I ask you to give such rationale right when posting the PR?

That was my intent. I agree, the commit message is hard to digest.

jerboaa:
Thanks, merged.

PR #41, opened 2017-01-31
Template to deploy dotnetcore 1.1 s2i

mglantz:
I've created a template to deploy dotnetcore 1.1 s2i apps. It includes asking for ASPNETCORE_URLS (required). Tested and works.

### DIFF
--- a/dotnet11-template.yaml
+++ b/dotnet11-template.yaml
@@ -1,0 +1,160 @@
+apiVersion: v1
+kind: Template
+labels:
+  template: dotnet11-s2i
+  dotnet: 1.1.0
+metadata:
+  annotations:
+    description: Application template for DOTNET applications built using S2I.
+    tags: dotnet,dotnet11
+    version: 1.1.0
+  name: dotnet11-s2i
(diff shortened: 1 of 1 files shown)


jerboaa:
[skipci]

jerboaa:
@mglantz Thanks for the contribution. I'm a bit unsure what problem this is trying to solve. For a couple of reasons:

1. There is already this "Try it" mechanism via the dotnet:1.1 imagestream, which seems to do the same thing. This is released.
2. There is this template for the to-be-released new sample app: https://github.com/redhat-developer/s2i-dotnetcore-ex/blob/dotnetcore-1.1/templates/aspnet-s2i-template.json Readme is here: https://github.com/redhat-developer/s2i-dotnetcore-ex/tree/dotnetcore-1.1

Are those not good enough? If so, why? Thanks!

mglantz:
@jerboaa Hey, I did not see those. I'll try em out for size :-)

jerboaa:
Closing for now. Please re-open if there is a use-case still not covered with the above.

PR #44, opened 2017-02-03
Revert a partial of 84a69e7 to listen port 8080.

tanaka-takayoshi:
Due to removing the ``UserUrls`` invocation deployed app can't be accessed as the application is listening port 8080 which is ASP.NET Core default whereas the container exposed port is 8080 (and the template provided by OpenShift 3.4 also presupposes listening port 8080).
I understand you'd like to make it flexible to change listening port. But the user can't deploy the working application with this repository by default. Working correctly by default setting should take priority over adding a new feature because this repository is now referred in OpenShift v3.4 web console.

### DIFF
--- a/1.0/test/asp-net-hello-world-envvar/src/app/Program.cs
+++ b/1.0/test/asp-net-hello-world-envvar/src/app/Program.cs
@@ -49,6 +49,7 @@ namespace SampleApp
             // Launch webserver
             var host = new WebHostBuilder()
                 .UseKestrel()
+                .UseUrls("http://0.0.0.0:8080")
                 .UseStartup<Startup>()
                 .Build();
 
--- a/1.0/test/asp-net-hello-world-envvar/src/app/Program.cs
+++ b/1.0/test/asp-net-hello-world-envvar/src/app/Program.cs
@@ -49,7 +49,6 @@ namespace SampleApp
             // Launch webserver
             var host = new WebHostBuilder()
                 .UseKestrel()
-                .UseUrls("http://0.0.0.0:8080")
                 .UseStartup<Startup>()
                 .Build();
 
--- a/1.0/test/asp-net-hello-world/Startup.cs
+++ b/1.0/test/asp-net-hello-world/Startup.cs
@@ -49,6 +49,7 @@ namespace SampleApp
                     options.NoDelay = true;
                     options.UseConnectionLogging();
                 })
+                .UseUrls("http://0.0.0.0:8080")
                 .UseContentRoot(Directory.GetCurrentDirectory())
                 .UseStartup<Startup>()
                 .Build();
--- a/1.1/test/asp-net-hello-world-envvar/src/app/Program.cs
+++ b/1.1/test/asp-net-hello-world-envvar/src/app/Program.cs
@@ -49,6 +49,7 @@ namespace SampleApp
             // Launch webserver
             var host = new WebHostBuilder()
                 .UseKestrel()
+                .UseUrls("http://0.0.0.0:8080")
                 .UseStartup<Startup>()
                 .Build();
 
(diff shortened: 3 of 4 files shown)


jerboaa:
[cibuild]

tmds:
@tanaka-takayoshi containers will now get the ASPNETCORE_URLS set to the 8080 port which makes the UseUrls unnecessary.

> Working correctly by default setting should take priority over adding a new feature because this repository is now referred in OpenShift v3.4 web console.

In what way is it referred in the OpenShift v3.4 web console? These are test projects to validate the s2i-dotnetcore functionality.

jerboaa:
@tanaka-takayoshi Does setting environment variable like so not work for you?

    ASPNETCORE_URLS=http://*:8080

This is what the [new image does](https://github.com/redhat-developer/s2i-dotnetcore/blob/master/1.1/Dockerfile.rhel7#L97). We also have a CI set up which verifies all builds and deploys. So this is somewhat surprising to us. @tmds your thoughts?

tanaka-takayoshi:
@tmds I'm Red Hat software maintenance engineer and tested on my own OpenShift v3.4 environment. When we create .NET app as the web console guide, buildconfig is like below.
```
  source:
    type: Git
    git:
      uri: 'https://github.com/redhat-developer/s2i-dotnetcore.git'
      ref: master
    contextDir: 1.0/test/asp-net-hello-world
``

Then the logs of deployed pod say:
```
Now listening on: http://localhost:5000
```

As far as I tested, ASPNETCORE_URLS is not seen in the application pod.

File: 1.1/test/asp-net-hello-world-envvar/src/app/Program.cs [hunk 1]
```diff
@@ -49,6 +49,7 @@ public static void Main(string[] args)
             // Launch webserver
             var host = new WebHostBuilder()
                 .UseKestrel()
+                .UseUrls("http://0.0.0.0:8080")
```
jerboaa: Please revert this as it is not used in the image stream as an example.

1.0/test/asp-net-hello-world/Startup.cs
```diff
@@ -49,6 +49,7 @@ public static void Main(string[] args)
                     options.NoDelay = true;
                     options.UseConnectionLogging();
                 })
+                .UseUrls("http://0.0.0.0:8080")
```
jerboaa: OK. We'll have to add this back in until our new image is released.

1.0/test/asp-net-hello-world-envvar/src/app/Program.cs
```diff
@@ -49,6 +49,7 @@ public static void Main(string[] args)
             // Launch webserver
             var host = new WebHostBuilder()
                 .UseKestrel()
+                .UseUrls("http://0.0.0.0:8080")
```
jerboaa: Please revert. Only `asp-net-hello-world` is the one referenced in the [current imagestream](https://github.com/redhat-developer/s2i-dotnetcore/blob/master/dotnet_imagestreams.json#L31).

tmds:
> When we create .NET app as the web console guide, buildconfig is like below.

The web console guide should not use test projects from this repository.

Can you try `76d5bea5e22713974edafa11bcf89b0be541aeea` instead of master?

jerboaa:
Thanks @tanaka-takayoshi for noticing this.

tmds:
@jerboaa the root cause is the use of `master` instead of a dedicated branch for the template. We have the same problem in the `s2i-dotnetcore-ex` repo.

tanaka-takayoshi:
Thank you, all.
I confirmed specifying the commit 76d5bea5e22713974edafa11bcf89b0be541aeea works well.
Also sets ASPNETCORE_URLS explicitly works well.
```
$ oc set env dc/<deploymentconfig> ASPNETCORE_URLS=http://*:8080
```

jerboaa:
@tmds Perhaps. Either way we need to make sure `asp-net-hello-world` in master works with released and unreleased images. The imagestreams have been loaded into OpenShift releases and if instantiated via the "Try it" mechanism they'll use master (we cannot unload what they've loaded, we'd need to do that to change to a specific revision). It shouldn't be too hard to keep that one example compatible.

jerboaa:
[cibuild]

PR #48, opened 2017-02-08
Remove DOTNET_RESTORE_ROOT

tmds:
Msbuild will be smart enough to figure out the dependent projects
it needs to restore to build DOTNET_STARTUP_PROJECT and DOTNET_TEST_PROJECTS.

For project.json DOTNET_RESTORE_ROOT allowed to control the projects
to restore.

Instead of introducing a new variable which will be deprecated
by msbuild, we won't introduce DOTNET_RESTORE_ROOT and instead
restore all project.json projects by doing a recursive restore.


jerboaa:
As far as I understand it this will limit the possibility to restore only a set of projects. Say app1 is within CWD, so is app2 and lib3. What if I wanted to restore only app2 and lib3 instead of all, app1, app2 and lib3? I suppose a user would have to override s2i scripts from our builder. Just making sure that's our take on it.

tmds:
> As far as I understand it this will limit the possibility to restore only a set of projects. Say app1 is within CWD, so is app2 and lib3. What if I wanted to restore only app2 and lib3 instead of all, app1, app2 and lib3? I suppose a user would have to override s2i scripts from our builder, right? Just making sure that's our take on it.

The user will no longer be able to only restore specific projects. If he really wants to he'll have to override the s2i script.

tmds:
@jerboaa can you merge this?

jerboaa:
[cibuild]

jerboaa:
Done, thanks!

PR #51, opened 2017-02-10
Run test projects before publish

tmds:
A publish could be a slow operation and tests should be fast.
So it makes more sense to run the tests before the publish
in order to make the build fail faster in case of failing tests.


tmds:
@jerboaa ptal

jerboaa:
Makes sense. I'll merge it in.

PR #55, opened 2017-02-14
Use s2i-dotnetcore-ex as primary sample app.

jerboaa:
Switch to redhat-developer/s2i-dotnetcore-ex as the primary sample
application shown in s2i-usage. The new sample app is a more exciting
example than then simple "Hello World" app we've used previously.

Updates to the image stream will happen later once the new images have
been released.

### DIFF
--- a/1.0/s2i/bin/usage
+++ b/1.0/s2i/bin/usage
@@ -9,7 +9,7 @@ To use it, install S2I: https://github.com/openshift/source-to-image
 
 Sample invocation:
 
-sudo s2i build https://github.com/redhat-developer/s2i-dotnetcore.git --context-dir=1.0/test/asp-net-hello-world/ ${NAMESPACE}/dotnetcore-10-${DISTRO}7 dotnet-sample-app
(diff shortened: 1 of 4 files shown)


jerboaa:
@tmds @omajid PTAL.

1.0/s2i/bin/usage
```diff
@@ -9,7 +9,7 @@ To use it, install S2I: https://github.com/openshift/source-to-image
 
 Sample invocation:
 
-sudo s2i build https://github.com/redhat-developer/s2i-dotnetcore.git --context-dir=1.0/test/asp-net-hello-world/ ${NAMESPACE}/dotnetcore-10-${DISTRO}7 dotnet-sample-app
+sudo s2i build https://github.com/redhat-developer/s2i-dotnetcore-ex.git --context-dir=app ${NAMESPACE}/dotnetcore-10-${DISTRO}7 dotnet-sample-app
```
tmds: Should there be a `--ref` parameter here, like in 1.1/.../usage?

jerboaa: It defaults to `--rev=master` so it's implicit.

tmds: Shouldn't 1.0 also have it's own branch?

jerboaa:
@tmds Done. More thoughts?

tmds:
:shipit:

PR #59, opened 2017-02-22
Run all tests in the test folder

tlbdk:
A normal dotnet project structure would be something along theses lines, so it would be nice to run all the tests in the test folder:

src/MyProject
src/MyProject.Integration
src/MyProject.Data
test/MyProject.Tests
test/MyProject.Integration.Tests
test/MyProject.Data.Tests


jerboaa:
OK, could you please add 1.0 equivalents and add a project under 1.1/test which has a shape like that? Otherwise there is a chance this will break in future. `test/run` runs those tests.

jerboaa:
[cibuild]

tlbdk:
I added the same code for 1.0 and copied asp-net-hello-world-envvar to  asp-net-hello-world-multitest since it has that structure already

jerboaa:
@tlbdk Thanks! The test script for 1.0/1.1 will need changing too so that it CI actually builds and runs the test project:
https://github.com/redhat-developer/s2i-dotnetcore/blob/master/1.0/test/run#L26

Just add `asp-net-hello-world-multitest` to the list.

tmds:
@tlbdk @jerboaa there is no reason we can assume that the user wants to run all tests in the test folder. These test can unintentionally slow down or even break the build when they can't run in OpenShift (e.g. because certain services are not available). To run all test projects under the test folder, the user can set DOTNET_TEST_PROJECTS=test/*.

jerboaa:
@tmds nice catch! I've just tried it and it works. So I don't think we need to get this merged.

@tlbdk See https://github.com/jerboaa/s2i-dotnetcore/blob/test_projects/1.1/test/asp-net-hello-world-envvar/.s2i/environment for an example which does exactly what you need.

jerboaa:
I'm closing this due to the `DOTNET_TEST_PROJECTS=test/*` environment which achieves something similar. The difference is that this PR would run the tests by default if DOTNET_TEST_PROJECTS is not set, but `test` exists. I'm sharing Tom's concern that not all tests might be automatically suitable to be run on the builder image, which might not have all the things required for running certain tests. For example, if a test would need an external program, `foo` say, which is not installed on the builder image the test is bound to fail. Being conservative here seems to be the better idea. I.e. allow for running all tests, but require the explicit env.

Issue #56, opened 2017-02-21
Add dotnet template


tmds:
Similar to https://github.com/redhat-developer/s2i-dotnetcore-ex/blob/master/templates/aspnet-s2i-template.json, but different...

- [ ] rename aspnet to dotnet

aspnet refers to the open source microsoft webstack. There are other popular webstacks like nancy (http://nancyfx.org/). s2i-dotnetcore is not 'aspnet' specific.

- [ ] add parameters for all s2i-dotnetcore environment variables

The names of these parameters match the environment names exactly.

- [ ] Store the template in this repo

The template is versioned with the rest of the s2i-dotnetcore repo.

@jerboaa @omajid what's do you think?

jerboaa:
+1. We should also massage the template to work independent of which dotnet core builder image is being used. Currently, the `dotnetcore-1.1` and `dotnetcore-1.0` branches have slightly different templates. This should be possible to achieve with parameters.

PR #60, opened 2017-02-24
add dotnet template

tmds:
The dotnet template can be used to create a new .NET Core service in OpenShift.
It provides parameters for all the environment variables of the s2i-dotnetcore
builder. It also includes a liveness and a readiness probe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,3 +45,17 @@ Image name structure
 3. Base builder image - rhel7
 
 Example: `dotnet/dotnetcore-10-rhel7`
+
+OpenShift Templates
+-------------------
+
+The `templates` folder contains OpenShift templates. Some of these will be shipped with OpenShift. If a template is not on your OpenShift installation, you can import it:
+
+```
+oc create -f <template.json>
+```
+
+**dotnet**
+
+The dotnet template can be used to create a new .NET Core service in OpenShift. It provides parameters for all the environment
+variables of the s2i-dotnetcore builder. It also includes a liveness and a readiness probe.
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ The `templates` folder contains OpenShift templates. Some of these will be shipp
 oc create -f <template.json>
 ```
 
-**dotnet**
+**dotnet-example**
 
-The dotnet template can be used to create a new .NET Core service in OpenShift. It provides parameters for all the environment
+The dotnet-example template can be used to create a new .NET Core service in OpenShift. It provides parameters for all the environment
 variables of the s2i-dotnetcore builder. It also includes a liveness and a readiness probe.
--- a/templates/dotnet-example.json
+++ b/templates/dotnet-example.json
@@ -201,7 +201,7 @@
                                         "scheme": "HTTP"
                                     },
                                     "initialDelaySeconds": 40,
-                                    "timeoutSeconds": 10
+                                    "timeoutSeconds": 15
                                 },
                                 "readinessProbe": {
                                     "httpGet": {
--- a/templates/dotnet-example.json
+++ b/templates/dotnet-example.json
@@ -109,7 +109,7 @@
                             },
                             {
                                 "name": "DOTNET_PUBLISH",
-                                "value": "Y"
+                                "value": "true"
                             }
                         ]
                     }
--- a/templates/dotnet-example.json
+++ b/templates/dotnet-example.json
@@ -275,7 +275,7 @@
         {
             "name": "APPLICATION_DOMAIN",
             "displayName": "Application Hostname",
-            "description": "The exposed hostname that will route to the Node.js service, if left blank a value will be defaulted.",
+            "description": "The exposed hostname that will route to the .NET Core service, if left blank a value will be defaulted.",
             "value": ""
         },
         {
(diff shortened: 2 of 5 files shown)


tmds:
This is implementing https://github.com/redhat-developer/s2i-dotnetcore/issues/56

templates/dotnet.json
```diff
@@ -0,0 +1,298 @@
+{
+    "kind": "Template",
+    "apiVersion": "v1",
+    "metadata": {
+        "name": "dotnet",
```
tmds: rename to "dotnet-example"

templates/dotnet.json
```diff
@@ -0,0 +1,298 @@
...
+        "annotations": {
+            "openshift.io/display-name": ".NET Core",
+            "description": "Template for creating .NET Core service",
```
tmds: update to: "An example .NET Core application"

templates/dotnet.json
```diff
@@ -0,0 +1,298 @@
...
+        }
+    ],
+    "parameters": [
+        {
+            "name": "NAME",
+            "displayName": "Name",
+            "description": "The name assigned to all of the frontend objects defined in this template.",
+            "required": true,
+            "value": "dotnet-service"
+        },
```
tmds: value: dotnet-example

README.md
```diff
@@ -45,3 +45,17 @@ Image name structure
...
+OpenShift Templates
+-------------------
+
+The `templates` folder contains OpenShift templates. Some of these will be shipped with OpenShift. If a template is not on your OpenShift installation, you can import it:
+
+```
+oc create -f <template.json>
+```
+
+**dotnet**
```
tmds: "dotnet-example", & update json filename

README.md
```diff
@@ -45,3 +45,17 @@ Image name structure
...
+
+The dotnet template can be used to create a new .NET Core service in OpenShift. It provides parameters for all the environment
```
tmds: The dotnet example ...

templates/dotnet.json
```diff
@@ -0,0 +1,298 @@
...
+        {
+            "name": "SOURCE_REPOSITORY_REF",
+            "displayName": "Git Reference",
+            "description": "Set this to a branch name, tag or other ref of your repository if you are not using the default branch."
+        },
+        {
+            "name": "CONTEXT_DIR",
+            "displayName": "Context Directory",
+            "description": "Set this to the relative path to your project if it is not in the root of your repository."
```
tmds: "Set this to use a subdirectory of the source code repository"

templates/dotnet.json
```diff
@@ -0,0 +1,298 @@
...
+            "name": "SOURCE_REPOSITORY_URL",
+            "displayName": "Git Repository URL",
+            "description": "The URL of the repository with your application source code.",
+            "required": true,
+            "value": "https://github.com/redhat-developer/s2i-dotnetcore-ex.git"
+        },
+        {
+            "name": "SOURCE_REPOSITORY_REF",
+            "displayName": "Git Reference",
+            "description": "Set this to a branch name, tag or other ref of your repository if you are not using the default branch."
```
jerboaa: This should perhaps default to "dotnetcore-1.0". That would match "dotnet:1.0" as the default builder image.

templates/dotnet.json
jerboaa: We should set this to "app" since that would allow one to instantiate the template with all the defaults, and it would build and run. s2i-dotnetcore-ex app sources are in "app"

tmds: I have a preference for setting DOTNET_STARTUP_PROJECT to demonstrate that parameter. I'd do the same for the dotnet+postgre template, if it wasn't for the big restore time of one of the projects we are not using. The postgre template can be update when we support msbuild.

templates/dotnet-example.json
```diff
@@ -0,0 +1,319 @@
...
+        },
+        {
+            "name": "CONTEXT_DIR",
+            "displayName": "Context Directory",
+            "description": "Set this to use a subdirectory of the source code repository"
+        },
+        {
+            "name": "APPLICATION_DOMAIN",
+            "displayName": "Application Hostname",
+            "description": "The exposed hostname that will route to the Node.js service, if left blank a value will be defaulted.",
```
jerboaa: I believe you don't want to reference Node.js do you?

templates/dotnet.json
jerboaa: OK, makes sense. I missed that.

templates/dotnet-example.json
tmds: Definitely not!

templates/dotnet-example.json
```diff
@@ -0,0 +1,324 @@
...
+                                "name": "DOTNET_TEST_PROJECTS",
+                                "value": "${DOTNET_TEST_PROJECTS}"
+                            },
+                            {
+                                "name": "DOTNET_CONFIGURATION",
+                                "value": "${DOTNET_CONFIGURATION}"
+                            },
+                            {
+                                "name": "DOTNET_PUBLISH",
+                                "value": "Y"
```
tmds: change to "true"

Issue #61, opened 2017-02-27
Control publish vs run


tmds:
Since the previous release, the builder will do a dotnet publish and invoke the application directly instead of build+run. This requires the publishOptions to be setup correctly.

We should avoid breaking applications that do not have their publishOptions setup correctly.

In case our next release is a 'breaking' one (msbuild) we should handle this as part of the documentation.

If the next release is still based on project.json, we should default to the old behavior and make this controllable via an envvar (e.g. DOTNET_PUBLISH=y). In the succeeding msbuild release, we can drop this flag.

PR #62, opened 2017-02-28
Add DOTNET_PUBLISH

tmds:
Used to control whether the application should be build by executing
`dotnet build` or `dotnet publish`. To publish the application set the
value to `Y`. It is recommended to publish your application. For
backwards compatibility, the default is ``. In the next major release
of the builder, this variable will be removed and the builder will
always publish the application.

### DIFF
--- a/1.0/README.md
+++ b/1.0/README.md
@@ -95,12 +95,12 @@ a `.s2i/environment` file inside your source code repository.
 
 * **DOTNET_PUBLISH**
 
-    Used to control whether the application should be build by executing
+    Used to control whether the application should be built by executing
     `dotnet build` or `dotnet publish`. To publish the application set the
-    value to `Y`. It is recommended to publish your application. For
-    backwards compatibility, the default is ``. In the next major release
-    of the builder, this variable will be removed and the builder will
-    always publish the application.
+    value to `true`. It is recommended to publish your application. For
+    backwards compatibility, the default is ``. In the next major release,
+    this variable will be removed and the builder will always publish the
+    application.
 
 * **DOTNET_ASSEMBLY_NAME**
 
--- a/1.0/README.md
+++ b/1.0/README.md
@@ -98,7 +98,7 @@ a `.s2i/environment` file inside your source code repository.
     Used to control whether the application should be built by executing
     `dotnet build` or `dotnet publish`. To publish the application set the
     value to `true`. It is recommended to publish your application. For
-    backwards compatibility, the default is ``. In the next major release,
+    backwards compatibility, the default is `false`. In the next major release,
     this variable will be removed and the builder will always publish the
     application.
 
--- a/1.0/README.md
+++ b/1.0/README.md
@@ -92,7 +92,15 @@ a `.s2i/environment` file inside your source code repository.
 
     Used to select the project to run. This must be the folder containing
     `project.json`. Defaults to `.`.
-    The application is built using the `dotnet publish` command.
+
+* **DOTNET_PUBLISH**
+
+    Used to control whether the application should be build by executing
+    `dotnet build` or `dotnet publish`. To publish the application set the
+    value to `Y`. It is recommended to publish your application. For
+    backwards compatibility, the default is ``. In the next major release
+    of the builder, this variable will be removed and the builder will
+    always publish the application.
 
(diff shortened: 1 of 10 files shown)


tmds:
Implements #61 

@jerboaa ptal

1.0/README.md
```diff
@@ -92,7 +92,15 @@ a `.s2i/environment` file inside your source code repository.
 
     Used to select the project to run. This must be the folder containing
     `project.json`. Defaults to `.`.
-    The application is built using the `dotnet publish` command.
+
+* **DOTNET_PUBLISH**
+
+    Used to control whether the application should be build by executing
```
jerboaa: Typo: "... should be built" not "build".

1.0/s2i/bin/assemble
```diff
@@ -54,22 +54,28 @@ for TEST_PROJECT in $DOTNET_TEST_PROJECTS; do
...
-
-# check if the assembly used by the script exists
-if [ ! -f "$DOTNET_PUBLISH_PATH/${APP_DLL_NAME}" ]; then
-  echo "error: Build output does not contain entrypoint assembly: ${APP_DLL_NAME}"
-  exit 1
-fi
-
-# Create run script in publish folder
-cat << EOF >"$DOTNET_RUN_SCRIPT"
+if [ "$DOTNET_PUBLISH" == "Y" ]; then
```
jerboaa: Could we change this to "true" over "Y". It would better represent that this is a boolean switch, IMHO.

1.0/test/qotd/project.json
```diff
@@ -15,11 +15,5 @@
     "netcoreapp1.0": {
       "imports": "dnxcore50"
     }
-  },
-  "publishOptions": {
-    "include": [
-      "quotes.txt",
-      "appsettings.json"
-    ]
```
jerboaa: Perhaps we should set DOTNET_PUBLISH=true for this test project so we can point users to an example which uses some files needed for publishing. Thoughts?

tmds: The envvar project has DOTNET_PUBLISH set and publishOptions. There was no point in removing this though, so I will add it back.

1.0/s2i/bin/assemble
tmds: true is ok for me, I'll update it

1.0/README.md
```diff
@@ -92,7 +92,15 @@ a `.s2i/environment` file inside your source code repository.
...
     Used to select the project to run. This must be the folder containing
     `project.json`. Defaults to `.`.
-    The application is built using the `dotnet publish` command.
+
+* **DOTNET_PUBLISH**
+
+    Used to control whether the application should be built by executing
+    `dotnet build` or `dotnet publish`. To publish the application set the
+    value to `true`. It is recommended to publish your application. For
+    backwards compatibility, the default is ``. In the next major release,
```
jerboaa: I suppose we can say the "... the default is `false`" rather than empty. Reads easier.

1.0/test/asp-net-hello-world-envvar/.s2i/environment
```diff
@@ -2,4 +2,5 @@ DOTNET_STARTUP_PROJECT=src/app
 DOTNET_CONFIGURATION=Debug
 DOTNET_TEST_PROJECTS=test/test1 test/test2
 DOTNET_ASSEMBLY_NAME=SampleApp
-DOTNET_RESTORE_SOURCES=https://api.nuget.org/v3/index.json https://www.myget.org/F/s2i-dotnetcore
\ No newline at end of file
+DOTNET_RESTORE_SOURCES=https://api.nuget.org/v3/index.json https://www.myget.org/F/s2i-dotnetcore
+DOTNET_PUBLISH=Y
```
jerboaa: This should be `true` not `Y`. That should fix the build.

jerboaa:
Looks good to me. Although I'd prefer `false` over `` default in docs :)

tmds:
@jerboaa let's hold off the merge until we know if the next release will be project.json/msbuild based

If it is msbuild, we can break and not merge this.

jerboaa:
@tmds OK.

PR #64, opened 2017-03-09
[README] Add a link to dotnet-example's sample usage.

jerboaa:
This depends on https://github.com/redhat-developer/s2i-dotnetcore-ex/pull/26

Rendered page here:
https://github.com/jerboaa/s2i-dotnetcore/tree/readme_reference_template_usage

[skipci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,5 +57,9 @@ oc create -f <template.json>
 
 **dotnet-example**
 
-The dotnet-example template can be used to create a new .NET Core service in OpenShift. It provides parameters for all the environment
-variables of the s2i-dotnetcore builder. It also includes a liveness and a readiness probe.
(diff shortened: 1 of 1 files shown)


tmds:
I added general template usage instructions to the dotnet-pgsql-persistent PR. They cover both templates. If they are ok for you, we can close this PR without merging.

jerboaa:
@tmds IMHO it would be good to have both. The template section of `s2i-dotnetcore-ex` talks in a bit more detail about how one would use the template to produce a 1.1 .NET Core build rather than 1.0 (the default). A crosslink does not hurt, does it?

tmds:
I prefer to have the usage instructions in the same repo as the templates and would avoid having similar instructions in various repositories.

jerboaa:
OK.

PR #69, opened 2017-04-20
[dotnet-pgsql-persistent] Use Secret objects for passwords.

jerboaa:
The OpenShift web console shows regular environment variables for
pods, et. al. However, for passwords or things that contain passwords
the web console should not show them by default. Using Secret objects
for secret values in the template will achieve this.

See also:
https://github.com/openshift/origin/pull/13528

Fixes RHBZ#1233513

Connect string after:
screenshot from 2017-04-20 15-01-56
Connect string before:
screenshot from 2017-04-20 15-03-20

PgSQL password after:
screenshot from 2017-04-20 15-02-24
PgSQL password before:
screenshot from 2017-04-20 15-03-47

### DIFF
--- a/templates/dotnet-pgsql-persistent.json
+++ b/templates/dotnet-pgsql-persistent.json
@@ -19,6 +19,26 @@
     },
     "objects": [
         {
+            "kind": "Secret",
+            "apiVersion": "v1",
+            "metadata": {
+                "name": "database"
+            },
+            "stringData": {
+                "database-password": "${DATABASE_PASSWORD}"
+            }
+        },
+        {
+            "kind": "Secret",
+            "apiVersion": "v1",
+            "metadata": {
+                "name": "database-connection"
+            },
+            "stringData": {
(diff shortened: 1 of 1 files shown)


jerboaa:
@tmds @bparees PTAL

templates/dotnet-pgsql-persistent.json
```diff
@@ -19,6 +19,26 @@
...
+        {
+            "kind": "Secret",
+            "apiVersion": "v1",
+            "metadata": {
+                "name": "database-connection"
+            },
+            "stringData": {
+                "connect-string": "Host=${DATABASE_SERVICE_NAME};Database=${DATABASE_NAME};Username=${DATABASE_USER};Password=${DATABASE_PASSWORD}"
+            }
+        },
```
bparees: you can use a single secret for all your keys:
https://github.com/openshift/origin/blob/master/examples/quickstarts/cakephp-mysql.json#L23-L34

jerboaa: OK thanks.

bparees:
lgtm

jerboaa:
Thanks, Ben!

Issue #13, opened 2016-12-07
DOTNET_CONFIGURATION needs to be tested


jerboaa:
Commit bca75a3daa09194bc85fe97e09d45911c9696a55 introduced two new environment variables which isn't covered by tests. We should fix this as we'd risk breaking them down the line.

jerboaa:
@omajid @gpiercey @tmds Anyone of you willing to work on those tests?

jerboaa:
https://github.com/redhat-developer/s2i-dotnetcore/pull/12 will test them, but it would still be good to add an explicit test case as to what is expected when DOTNET_CONFIGURATION=Debug vs. DOTNET_CONFIGURATION=Release. Similarly, the use case for  DOTNET_FRAMEWORK should be illustrated in test.

jerboaa:
I'm not sure DOTNET_FRAMEWORK works correctly, when I try test/run on the 1.0 folder I'm getting:
```
I1207 13:09:36.838666 14271 environment.go:60] Setting 4 environment variables provided by environment file in sources
---> Copying application source ...
---> Installing dependencies ...
log  : Restoring packages for /opt/app-root/src/src/lib/project.json...
log  : Writing lock file to disk. Path: /opt/app-root/src/src/lib/project.lock.json
log  : src/lib/project.json
log  : Restore completed in 545ms.
log  : Restoring packages for /opt/app-root/src/src/app/project.json...
log  : Installing System.Diagnostics.Contracts 4.0.1.
log  : Installing System.Text.Encodings.Web 4.0.0.
log  : Installing System.Net.WebSockets 4.0.0.
log  : Installing Microsoft.Net.Http.Headers 1.0.0.
log  : Installing Microsoft.AspNetCore.WebUtilities 1.0.0.
log  : Installing Microsoft.Extensions.ObjectPool 1.0.0.
log  : Installing Microsoft.Extensions.FileSystemGlobbing 1.0.0.
log  : Installing Microsoft.AspNetCore.Http.Abstractions 1.0.0.
log  : Installing Microsoft.Extensions.FileProviders.Abstractions 1.0.0.
log  : Installing Microsoft.AspNetCore.Http.Features 1.0.0.
log  : Installing Microsoft.Extensions.DependencyInjection.Abstractions 1.0.0.
log  : Installing Microsoft.Extensions.Logging.Console 1.0.0.
log  : Installing Microsoft.AspNetCore.Http.Extensions 1.0.0.
[...]

PR #72, opened 2017-05-02
Added tests for DOTNET_CONFIGURATION environment variable.

aslicerh:
This addresses redhat-developer/s2i-dotnetcore#13


tmds:
LGTM.
nit: Perhaps you can drop the envvar-general folder and put the projects at the same level?
Can you add this test to the 2.0 branch? (separate PR)

aslicerh:
@tmds Yeah, I was just concerned that we might end up with a lot of folders that test environment variables and that it might clutter up the test projects. That said, maybe we should fix that once it becomes an issue and not before. I'll make that change.

PR #73, opened 2017-05-02
Restore and publish using the rhel7 rid.

tmds:
This publishes an app for rhel7 that uses the shared runtime and that
doesn't include dependencies needed to run on the shared runtime of
other platforms (e.g. osx, windows, ubuntu).

### DIFF
--- a/2.0/s2i/bin/assemble
+++ b/2.0/s2i/bin/assemble
@@ -8,6 +8,7 @@ DOTNET_STARTUP_PROJECT="${DOTNET_STARTUP_PROJECT:-.}"
 
 # Private environment
 DOTNET_FRAMEWORK="netcoreapp2.0"
+DOTNET_RID=rhel.7-x64
 
 # npm
 if [ -n "${DOTNET_NPM_TOOLS}" ]; then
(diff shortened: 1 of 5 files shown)


2.0/s2i/bin/assemble
```diff
@@ -52,26 +53,25 @@ else
...
 for SOURCE in $DOTNET_RESTORE_SOURCES; do
   RESTORE_OPTIONS="$RESTORE_OPTIONS --source $SOURCE"
 done
 
 # run tests
 for TEST_PROJECT in $DOTNET_TEST_PROJECTS; do
+    echo "---> Restoring test project ($TEST_PROJECT) dependencies..."
+    dotnet restore "$TEST_PROJECT" -r "$DOTNET_RID" $RESTORE_OPTIONS
     echo "---> Running test project: $TEST_PROJECT..."
-    pushd "$TEST_PROJECT"
```
tmds: I added these in the initial 2.0 to make the dotnet commands to look for the project in the current dir. Now, I replaced this by passing the project explicitly to the dotnet commands.

aslicerh: Cool. Just curious.

PR #75, opened 2017-05-03
DOTNET_STARTUP_PROJECT and DOTNET_TEST_PROJECTS can be csproj files

tmds:


tmds:
[cibuild]

tmds:
I don't get the Jenkins build log.
It's complaining about pushd, but that has been removed in the previous commit.

> E0503 07:56:50.585737 15347 util.go:91] /usr/libexec/s2i/assemble: line 63: pushd: test/test1/test1.csproj: Not a directory

aslicerh:
This may be off, but is the assemble file at /usr/libexec/s2i/assemble the one from this branch or is it a pre-installed (or older) version?

tmds:
[cibuild]

tmds:
@aslicerh it should use the version that is on the branch. I just saw the jenkins config is responsible for deleting the images and added 20.

PR #81, opened 2017-05-22
Disable unneeded yum repositories during build.

aslicerh:


2.0/Dockerfile.rhel7
```diff
@@ -42,7 +42,8 @@ RUN yum repolist --disablerepo=* && \
     yum-config-manager --enable rhel-7-server-rpms > /dev/null
```
jerboaa: Lines 39-42 can probably be removed now.

aslicerh: PR #82 removes those lines and makes a comment about this change being more permanent.

PR #83, opened 2017-05-30
[WIP] Create a runtime image from a build image

aslicerh:
[Bug 1430280](https://bugzilla.redhat.com/show_bug.cgi?id=1430280) relates to having a seperate runtime image that only contains the necessary runtime dependencies for a project, but the build dependencies. This is to reduce the size of the container image that actually runs the application.

With this you can point the build_runtime_image.sh script to a build image (that was built using s2i) and generate a runtime image.

Building the runtime on a rhel7 image brings the image size down from a little over 2 GB to about 380 MB. About 2/3 of what is left is the base OS image and the dotnet runtime.

Currently, there are no automated tests for it, and it needs a lot of safety checks. It also needs the correct documentation headers and updates to the project readme files related to it.

### DIFF
--- a/2.0/build-runtime_image.sh
+++ b/2.0/build-runtime_image.sh
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+# Setup initial variables
+RUNTIME_BASE_IMAGE=${RUNTIME_BASE_IMAGE:-rhel7}
+BUILD_IMAGE=$1
+BINARY_TAR=publish.tar.gz
+DOCKERFILE_TEMPLATE=${DOCKERFILE_TEMPLATE:-runtime-dockerfile-template}
+# TODO: is here a way to get this from the build image?
+DOTNET_PUBLISH_PATH=${DOTNET_PUBLISH_PATH:-/opt/app-root/publish}
+
+# Check for missing files and images to fail early
+if ! $(docker inspect ${BUILD_IMAGE} &> /dev/null); then
+	echo "Could not locate a docker image by the name of '${BUILD_IMAGE}'."
+	exit 1
+fi
+if ! $(docker inspect ${BUILD_IMAGE} &> /dev/null); then
+	echo "Could not locate a docker image by the name of '${RUNTIME_BASE_IMAGE}'."
+	exit 1
+fi
+if [ ! -e ${DOCKERFILE_TEMPLATE} ]; then
+	echo "Could not locate runtime dockerfile file template."
+	echo "It was expected to be '${DOCKERFILE_TEMPLATE}'"
+	exit 1
+fi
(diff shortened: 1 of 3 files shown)


jerboaa:
@aslicerh I'm a bit confused. If I understand [this doc](https://docs.openshift.org/latest/dev_guide/builds/advanced_build_operations.html#dev-guide-chaining-builds) correctly, there should be 3 ingredients to an chained build approach:

1. s2i builder image (`dotnet/dotnetcore-1x-rhel7` currently for us)
2. a pre-built runtime image with just .NET core in it
3. A Dockerfile which is used to extract the built artifacts from a build of sources plus image in 1.

Can you explain how your `build_runtime_image.sh` relates to such a setup? Would it be something to replace step 3 with? Would we not need a base runtime image too? Step 2. above?

I expected us needing to implement a new .NET runtime base image, call it `dotnet/dotnet-20-runtime`, which our s2i-dotnetcore builder image should probably use in "FROM" and add in the bits it needs only for building purposes (nuget, node, etc). Then we'd have 1. and 2, and would be missing only 3. Am I wrong?

aslicerh:
@jerboaa Just to start, this script is pretty much the '3' is your list, since the '2' is mostly trivial, as I'll try to point out below.

What exactly would a 'dotnet runtime' image be? Without having any insight into the base project that needs to be run, all it could really be would be something like rhel7 with dotnet (not the sdk) installed on it. You don't really need a prebuilt image for that, just a line in the dockerfile to install that via yum. The current dockerfile template here downloads the whole sdk and removes the sdk part, but that is because we don't have a 2.0 runtime only package yet.

When it does come to copying the build artifacts over to a runtime image, you can't really just use a dockerfile. The only way to access stuff from one image into a new image is to use a FROM. However, that pulls in the entirety of that image, so if you FROM a build image (to get the build artifacts) then your runtime image is identical to the build image.

While it is true that you can copy files to/from the host to an image in a dockerfile, you can not access files in an image without spinning up a container with said image, and I do not believe that you can do that from a dockerfile.

So what you need is to spin up a container of the build image, copy the build artifacts (the publish directory) to the host and then copy that into the runtime image. That is what this script does and it is also what openshift does in the link you pointed to.

So we could build an rh-dotnet image that already has just dotnet (not the sdk) installed. But I'm not sure what value it gives to the consumer, other than a slightly reduced image creation time. In the mean time, by adding the dotnet install as part of the script here (instead of putting it into a base image) it means that you can use any yum-compatible image as your base image (including the stripped down atomic images we have been talking about, or a fedora/centos image).

aslicerh:
Closing this as not quite relevant in it's current form.

PR #84, opened 2017-06-05
Create runtime and build images.

aslicerh:
Existing files moved from \<ver\>/\<stuff\> to \<ver\>/build/\<stuff\>.
Files for building runtime images can be found at \<ver\>/runtime/\<stuff\>.
Test files are still located at \<ver\>/test.

As part of this the test script needed to be reworked a bit so that it didn't assume only one current and valid image.


2.0/runtime/Dockerfile.rhel7
```diff
@@ -0,0 +1,42 @@
...
+# Don't do initially populate of package cache
+ENV DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
+
+## By default, ASP.NET Core runs on port 5000. We configure it to match
+## the container port.
+ENV ASPNETCORE_URLS=http://*:8080
+
+# TODO: remove upstream sdk
+# install upstream 2.0 sdk
+RUN yum install --disablerepo=\* --enablerepo=rhel-7-server-rpms -y libunwind libicu && \
```
jerboaa: We should probably install `nss_wrapper` and `tar` in the runtime image (i.e. move it away from the build image). `nss_wrapper` is the tool being used for username look-ups for random UIDs. See https://github.com/redhat-developer/s2i-dotnetcore/blob/master/1.1/s2i/bin/run#L2.

build.sh
```diff
@@ -43,21 +43,38 @@ check_result_msg() {
...
-      docker build -f Dockerfile.rhel7 -t ${img_name} .
-      check_result_msg $? "Building Docker image ${img_name} FAILED!"
-    fi
-    echo "Running tests on image ${img_name} ..."
-    ./test/run
-    check_result_msg $? "Tests for image ${img_name} FAILED!"
-  popd &>/dev/null
+  build_name="dotnet/$(image_name ${v})"
+  runtime_name="${build_name}-runtime"
+  build_image "${v}/runtime" "${runtime_name}"
```
jerboaa: Could you add a comment "Build the runtime image" or some such. It took me a while to grok what's actually building the runtime image and what the builder image. Same for the builder image.

tmds: Can we move the build stuff into a per version build script?
For the 1.x versions there is no need to add a runtime, build folder.
And when developing on a specific version, we can just call those scripts.

2.0/test/run
```diff
@@ -429,11 +504,39 @@ test_new_web_app() {
...
 }
 
+create_runtime_app() {
+  # Create a temp directory to hold our temporary files
+  TMPDIR=$(mktemp -d)
+  # TODO: Figure out a better way to allow access from the image /OTHER/ than 777
+  chmod 777 ${TMPDIR}
+  echo "Created temp directory ${TMPDIR}"
+  # Grab the build artifacts from the build image
+  docker run --rm --volume=${TMPDIR}:/opt/app-root/tmp:Z "${IMAGE_NAME}-testapp" tar zcvpf /opt/app-root/tmp/publish.tar.gz -C /opt/app-root/publish .
```
tmds: What is the significance of `/opt/app-root/tmp`? Is this a convention?

2.0/test/run
```diff
@@ -429,11 +504,39 @@ test_new_web_app() {
...
   check_result $?
 
-  cleanup_app
+  cleanup_container "${cid_file}"
   info "All tests for the ${app} finished successfully."
   cleanup ${app}
 }
 
+create_runtime_app() {
+  # Create a temp directory to hold our temporary files
```
tmds: how does openshift know what runtime image to use for a certain sdk image?

2.0/build/Dockerfile.rhel7
```diff
@@ -1,4 +1,4 @@
-FROM rhel7
+FROM dotnet/dotnet-20-rhel7-runtime
 # This image provides a .NET Core 2.0 environment you can use to run your .NET
 # applications.
```
tmds: Remove EXPOSE and ENVs set in the base image.

2.0/runtime/Dockerfile.rhel7
```diff
@@ -0,0 +1,42 @@
...
+# applications.
+
+EXPOSE 8080
+
+ENV DOTNET_CORE_VERSION=2.0
+
+# Default to UTF-8 file.encoding
+ENV LANG=C.UTF-8 \
+    HOME=/opt/app-root \
+    PATH=/opt/app-root/src/.local/bin:/opt/app-root/src/bin:/opt/app-root/bin:/opt/app-root/node_modules/.bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
tmds: Don't include the node paths here, should be in the sdk image.

2.0/build/Dockerfile.rhel7
tmds: How does DOTNET_PUBLISH_PATH, DOTNET_RUN_SCRIPT fit in here (and with /opt/app-root/tmp)?

2.0/runtime/Dockerfile.rhel7
aslicerh: Noted, and will do. And yeah, I figured there would be a bit of shuffle of various env's and packages between the two dockerfiles, so feel free to point out anything that you find questionable.

2.0/test/run
aslicerh: I'm mounting a host directory into the container. The host directory being mounted is the temp dir we just created. The part after the colon (/opt/app-root/tmp) is the container directory it gets mounted to (the second colon with the Z is to tell docker to relabel the directory for SELinux support).

This while bit is to copy the build artifacts from the build image to the local host. We do this since you can't copy directly image to image.

And no, as far as I am aware, '/opt/app-root/tmp' is not any real convention, but it only exists for the brief time it takes to copy the runtime artifacts and does not make it into the final image or container.

2.0/test/run
aslicerh: If you want to use the runtime image to deploy your app on then you need a build config file that does two things:
1) it specifies the runtime image as its source.
2) it specifies a source path that locats the artifact zip on the build image.

I should have a PR up for a sample config file in s2i-dotnetcore-ex doing this fairly soon.

build.sh
aslicerh: I don't know, it seems like the scripts would be mostly identical, just with slightly different file paths. I could change the build.sh script to know which path to use based on the version number.

I could also make 1.0 and 1.1 also build runtime images, since the only thing really needed is the new dockerfile.

2.0/build/Dockerfile.rhel7
aslicerh: Ok, so a bunch of these ENV's pertain to s2i specific stuff. But there is nothing about the runtime image that is s2i specific. Right now it doesn't even have the s2i files on it. Since the runtime image can be used with non-s2i builds (either by hand or via OpenShift) should we put s2i stuff in it?

2.0/test/run
```diff
@@ -429,11 +504,39 @@ test_new_web_app() {
...
+  echo "Created temp directory ${TMPDIR}"
+  # Grab the build artifacts from the build image
+  docker run --rm --volume=${TMPDIR}:/opt/app-root/tmp:Z "${IMAGE_NAME}-testapp" tar zcvpf /opt/app-root/tmp/publish.tar.gz -C /opt/app-root/publish .
+
+  if [ ! -e "${TMPDIR}/publish.tar.gz" ] || [ ! -s "${TMPDIR}/publish.tar.gz" ]; then
+    echo "Failed to extract the publish directory from the build image '${app}'."
+    cleanup
+    exit 1
+  fi
+  echo "FROM ${RUNTIME_IMAGE_NAME}" > "${TMPDIR}/Dockerfile"
```
aslicerh: Yeah, my only issue with that is that he EOF can't have any leading whitespace, and that messes up the indentation of the subroutine here. That's why I went with the echo. Though do not particularly mind one way or the other. I just defaulted to 'preserve readability at the cost of boilerplate'.

build.sh
aslicerh: Done.

2.0/runtime/Dockerfile.rhel7
aslicerh: Removed. I also removed the pointers to 'opt/app-root/src/*'

build.sh
tmds: There is a lot of identical stuff in the version folders. My preference is to have a build script per folder. We don't want runtime images for 1.x.

2.0/test/run
tmds: I don't understand why the artifacts end up at /opt/app-root/tmp in the build image.

2.0/test/run
tmds: It's unfortunate the builder image can't tell open shift what runtime image to use and where it has placed the output content.

2.0/build/Dockerfile.rhel7
tmds: The builder image is setting stuff like LANG, HOME, PATH, ...
Doesn't the runtime image need an s2i run script?

2.0/test/run
aslicerh: They don't. Here, lets break this line down:
docker run --rm => run an image in a container and remove that container when it is done

--volume=${TMPDIR}:/opt/app-root/tmp:Z => Mount $TMPDIR from the local host to the 'opt/app-root/tmp' directory on the container

"${IMAGE_NAME}-testapp" => Use this image to build the container

tar zcvpf /opt/app-root/tmp/publish.tar.gz -C /opt/app-root/publish . => cd into /opt/app-root/publish, create/tar/compress everything in it, and output the archive to /opt/app-root/tmp/publish.tar.gz

So what this is doing is building a container and running a tar command inside the publish directory and then copying the tar file to the tmp directory.

2.0/test/run
aslicerh: My first PR was a script to build a complete runtime image outside of OpenShift. You could use that script to build said image and use it inside of OpenShift. But doing it this way (using the custom build config) you do get to use OpenShift triggers, so the images will automatically update as patches and things come up.

So yeah, there are pros and cons to both approaches.

tmds: I meant something else: OpenShift could have a feature that allows a builder image to specify where it puts the artifacts and what runtime image is appropriate.

2.0/build/Dockerfile.rhel7
aslicerh: The runtime only needs an s2i run script if your app needs an s2i run script. And if it does then it should be copied over as part of the build artifacts. Which is exactly what happens right now in the tests.

Remember, this runtime image is created before the build image is. In fact, it is a dependency for the build image. So (as it exists right now) there are no requirements placed on the runtime image based on your app. All of the app required stuff should be done when you build the runtime image for your app.

Maybe this will help. This is a rough mapping of the build process idea:
```
base runtime image ----> base build image ----> build image for app (with build artifacts)
                    \                                                                      \
                      \----------------------------------------------------------------------------> runtime image for app (with build artifacts copied from build image for app)

File: 2.0/test/run (hunk 3 above)
Comment (tmds) [inline 2.0/test/run]: Got it.

File: 2.0/build/Dockerfile.rhel7 (hunk 5 above)
Comment (tmds) [inline 2.0/build/Dockerfile.rhel7]: So the runtime image also figures out the s2i run script based on the build config? Looking forward to see the build config example...
Anyhow, take a look at the envvars and remove stuff from the base build image that gets inherited from the base runtime image (like the LANG, HOME, PATH, EXPOSE).

Comment (aslicerh) [inline 2.0/build/Dockerfile.rhel7]: You need a dockerfile to build the app's runtime image. In that dockerfile is where you'll specify stuff like your app specific env vars, or you apps run cmd (which may be s2i run, or a usage script).

If you look in the tests you'll see I have a generic one declared inline in the test script.

OpenShift also expects you to use a Dockerfile for the runtime image and the examples all give it inline with the custom build config file.

File: 2.0/test/run (hunk 7 above)
Comment (jerboaa) [inline 2.0/test/run]: Right. Up to you, really. I find the cat syntax more readable since it shows the entire dockerfile without much clutter.

File: build.sh (hunk 2 above)
Comment (jerboaa) [inline build.sh]: I'm in preference to having only one top-level build script. Folders with versions should basically be `docker build -f <Dockerfile> -t <sometag>`. Then run tests via `./test/run.sh`. On that note, we should have a subset of tests we run on just the runtime image. I.e. testing of the runtime image should not depend on the builder image. So maybe the folder structure should be this:

    <version>
         |- runtime
         |        |-- test
         |         `- Dockerfile.rhel7
          `- build
                 |-- test
                  `- Dockerfile.rhel7

That structure would keep tests/image files self-contained.

Comment (aslicerh) [inline build.sh]: I thought about that, but there would be so much copied in the runtime test script from the build test script.

The current test script has different tests it runs for build and runtime, though I think they could be better separated for one building (and test) one of the images at a time.

File: 2.0/test/run (hunk 3 above)
Comment (jerboaa) [inline 2.0/test/run]: Maybe we could use a more meaningful name such as `/opt/app-root/build-img-mount` or some such? I'm considering a case where this dir sticks around for some reason. Having `tmp` in there is less telling than `build-img-mount`. My $0.02

Comment (aslicerh) [inline 2.0/test/run]: Sure, we can make the name more meaningful. I mostly didn't bother because it only exists long enough to run this single tar command.

File: 2.0/build/Dockerfile.rhel7 (hunk 5 above)
Comment (tmds) [inline 2.0/build/Dockerfile.rhel7]: I see. This is a lot less user friendly than I assumed it would be.

Comment (aslicerh) [inline 2.0/build/Dockerfile.rhel7]: It is less user friendly than I would like for it to be.

If we assume the the app is going to use s2i (which would make sense, since this is in the s2i project), then we can pre-configure a lot of this stuff for the user. They would still need a custom openshift build config, but that is going to be necessary no matter what we do (because of how openshift does chained builds).

The downside to assuming s2i is that this same runtime image is going to be used to build the pre-built runtime containers, and there is no guarantee that a customer would even know about this project, much less care about the s2i stuff in it.

This case differs somewhat from the provided build containers because the build containers are kind of (but not quite) required to use s2i (or at least it is recommended in the usage). But a customer may be building their project by hand here and just wanting an image to run it on.

File: build.sh (hunk 2 above)
Comment (jerboaa) [inline build.sh]: Use your good judgement. My worry was downstream syncs. If you think it's OK with this setup then fine with me. For runtime image tests we could upload pre-built binaries, fwiw.

File: 2.0/build/Dockerfile.rhel7 (hunk 5 above)
Comment (aslicerh) [inline 2.0/build/Dockerfile.rhel7]: Does this question still need to be addressed, given our other discussions?

**Comment by aslicerh** [comment]
Ok, I think I've addressed everything commented on so far (though the README may still need some tweaking depending on what we want out of it).

File: 2.0/runtime/README.md [hunk 8]
```diff
@@ -0,0 +1,71 @@
...
+---------------------
+
+* **ASPNETCORE_URLS**
+
+    This variable is set to `http://*:8080` to configure ASP.NET Core to use the
+    port exposed by the image.
+
+* **HTTP_PROXY, HTTPS_PROXY**
+
+    Configures the HTTP/HTTPS proxy used when building and running the application.
```
jerboaa: Since this is a runtime image, I suppose we should drop "building and" here. I.e. "Configures htt HTTP/HTTPS proxy used when running the application"

2.0/runtime/test/run
```diff
@@ -0,0 +1,450 @@
...
+}
+
+create_runtime_app() {
+  local app="$1"
+  # Create a temp directory to hold our temporary files
+  TMPDIR=$(mktemp -d)
+  # TODO: Figure out a better way to allow access from the image /OTHER/ than 777
+  chmod 777 ${TMPDIR}
+  echo "Created temp directory ${TMPDIR}"
+  cp "${app}.tar.gz" "${TMPDIR}/${app}.tar.gz"
```
jerboaa: Perhaps it would make sense to add to the README how those binary apps need to get created, in case some app code needs changing.

2.0/build/Dockerfile.rhel7
```diff
@@ -1,10 +1,11 @@
...
 
 EXPOSE 8080
```
tmds: can you remove the expose?

2.0/build/Dockerfile.rhel7
```diff
@@ -1,10 +1,11 @@
...
 
 ENV DOTNET_CORE_VERSION=2.0
```
tmds: same: remove DOTNET_CORE_VERSION

2.0/build/Dockerfile.rhel7
```diff
@@ -1,10 +1,11 @@
...
+
 # Default to UTF-8 file.encoding
 ENV LANG=C.UTF-8 \
```
tmds: remove: LANG

2.0/build/Dockerfile.rhel7
```diff
@@ -1,10 +1,11 @@
...
     HOME=/opt/app-root \
```
tmds: remove: setting HOME

tmds: Can you set PATH by adding to $PATH from the base image?

2.0/build/Dockerfile.rhel7
```diff
@@ -52,11 +53,9 @@ RUN yum install --disablerepo=\* --enablerepo=rhel-7-server-rpms -y yum-utils &&
 
 # TODO: remove upstream sdk
 # install upstream 2.0 sdk
-RUN yum install -y libunwind libicu && \
-    curl https://download.microsoft.com/download/0/6/5/0656B047-5F2F-4281-A851-F30776F8616D/dotnet-dev-linux-x64.2.0.0-preview1-005977.tar.gz -o /tmp/dotnet-sdk.tar.gz && \
-    mkdir -p /opt/dotnet && \
-    tar xf /tmp/dotnet-sdk.tar.gz -C /opt/dotnet && \
-    ln -s /opt/dotnet/dotnet /usr/bin/dotnet
+RUN curl https://download.microsoft.com/download/0/6/5/0656B047-5F2F-4281-A851-F30776F8616D/dotnet-dev-linux-x64.2.0.0-preview1-005977.tar.gz -o /tmp/dotnet-sdk.tar.gz && \
```
tmds: This is also in the base image, so it can be removed here.

2.0/build/Dockerfile.rhel7
```diff
@@ -52,11 +53,9 @@ RUN yum install --disablerepo=\* --enablerepo=rhel-7-server-rpms -y yum-utils &&
...
+	tar xf /tmp/dotnet-sdk.tar.gz -C /opt/dotnet "./sdk" && \
+	rm /tmp/dotnet-sdk.tar.gz
 
 # Switch to default app-user for cache population
 USER 1001
```
tmds: Set NUGET_XMLDOC_MODE=skip in the base image
And remove NUGET_XMLDOC_MODE, DOTNET_SKIP_FIRST_TIME_EXPERIENCE from the sdk image

tmds: Remove ASPNETCORE_URLS too

tmds:
Q: when doing a split build, is HTTP(S)_PROXY still set in the runtime image?
Q: can you document how to use this in OpenShift?

2.0/build/Dockerfile.rhel7
aslicerh: Ok, so regarding removing everything in the runtime image from the build image.

There is nothing (as of this moment) that requires the build image to be based off of the runtime images, except that we want to make sure that the dotnet versions installed are locked together when we build containers. 

So if you just wanted to build the build image without the runtime, you just change the 'FROM' in the dockerfile and now you have one built off of Fedora or Centos. If we remove all of those elements then you either have to copy stuff from the runtime image (and figure out what all you need to copy) or you need to always build both and base the runtime image off of Fedora or Centos.

Now, given that I, personally, don't mind always building both I don't mind making these changes. But I thought I would be explicit about what that might impact.

jerboaa: Yes, building both is fine. For centos mirror images, everything should be the same. One runtime one build image. Image hierarchy being the same etc.

2.0/build/Dockerfile.rhel7
aslicerh: This is a temporary thing until the official 2.0 packages get released.

For the runtime image, this downloads the zip containing the binaries and the sdk, and removed the sdk. For the build image it downloads it again and just extracts the sdk.

In the final dockerfiles this will just be a yum command. The runtime will install the dotnet runtime package, the build image will install the dotnet sdk package.

tmds: I'd minimize the temporary code by removing it here and keeping the sdk in the base image. Up to you.

aslicerh: For this one I think I'll leave it as is. Leaving the sdk around would negate most of the 'size savings' from the runtime image, while still needing both dockerfiles to be changed at the end. Its current form most approximates how the actual images will look, which also will make any testing more accurate.

aslicerh:
@tmds 
> Q: when doing a split build, is HTTP(S)_PROXY still set in the runtime image?
That is a good catch. Normally this variable is set via s2i and becomes a built in env var for the app image. But clearly that setting won't get magically transferred into the runtime image.

The user could define the env var in the dockerfile that is used to build the actual runtime image. But that is a horrible user experience compared to the standard s2i usage. Let me do some research on how openshift handles this.

> Q: can you document how to use this in OpenShift?
Can you clarify your question? Are asking if we will put an openshift example, say, in the readme of the runtime image? Or is this a more general question of if one will be provided at all?

tmds:
For the HTTP(S)_PROXY, we could inject these variables in the run script we generate, if openshift doesn't do that.
Related to this, I wonder if OpenShift supports setting environment variables at a larger scope (e.g. at project level). It seems tedious to have to add these to every build separately.

For the doc, update README. Perhaps things will be most clear if we have a single README that describes both images and how they relate.

jerboaa:
@tmds FWIW, Openshift supports setting environment variables via `oc env`. For the HTTP(S)_PROXY variables this should be enough for the runtime image. .NET will pick it up. For the builder, it needs to use those settings for npm explicitly so we needed to do some action (e.g. `npm config set proxy`).

tmds:
> @tmds FWIW, Openshift supports setting environment variables via oc env

It doesn't seem possible to set this at project scope to apply this to all containers.

jerboaa:
> It doesn't seem possible to set this at project scope to apply this to all containers.

Right. Wouldn't templates with parameters address this somewhat? It seems an OpenShift issue, rather than a .NET container issue.

tmds:
> Right. Wouldn't templates with parameters address this somewhat? It seems an OpenShift issue, rather than a .NET container issue.

Yea. It's an openshift issue. HTTP(S)_PROXY are general global. So made me wonder how one would typically set this.
Suppose they are set globally and we'd decide to inject them in the s2i run script. Then when the global setting changes, application images would still be using the old value. Which would be a bug.

aslicerh:
Ok, I think this should be it.

2.0/runtime/README.md
```diff
@@ -68,4 +84,4 @@ Environment variables
 
 * **HTTP_PROXY, HTTPS_PROXY**
```
tmds: The **HTTP_PROXY, HTTPS_PROXY** section can be removed since the runtime image doesn't do anything special with them.

2.0/runtime/README.md
```diff
@@ -14,12 +14,28 @@ example, given an archive of the binaries of a simple [dotnet-sample-app](test/a
 application, the following Dockerfile could be used.
 
 **Sample Dockerfile**
+```
```
tmds: This isn't the example I was looking for.
How to use this in OpenShift? i.e. template description.

2.0/runtime/README.md
```diff
@@ -14,12 +14,28 @@ example, given an archive of the binaries of a simple [dotnet-sample-app](test/a
...
 COPY asp-net-hello-world.tar.gz asp-net-hello-world.tar.gz
 RUN tar xvf asp-net-hello-world.tar.gz
 CMD /opt/app-root/publish/s2i_run
+```
+
+As for generating the binary archives in the first place, often this can be
+accomplished by running a dotnet restore and a dotnet publish to a single
+directory and then archiving that directories contents.
+
+**Example Binary Archive Creation**
```
tmds: I don't think this is a meaningful example

aslicerh: This was added in response to something @jerboaa asked for, not in regards to your openshift example request.

2.0/runtime/README.md
aslicerh: There is going to be an example template added to the s2i-dotnetcore-ex repo.

Do you still want an example added to the README here?

2.0/runtime/README.md
aslicerh: If you change the environment variable in the runtime image that change should carry over to any subsequent images. So a runtime image built with the HTTP_PROXY setting configured should still have that configuration for any application that depends on the runtime image. It's also the only setting in the build image (other than ASPNETCORE_URLS) that still has some relevance outside of build time for that reason, hence why I kept it here.

But if you want it gone I'll take it out.

2.0/runtime/README.md
tmds: > Can we add the example template, to the templates folder in this repo?

Definitely. From the README it should be clear how the runtime image can be used as a target of the sdk image in OpenShift.

> There is going to be an example template added to the s2i-dotnetcore-ex repo.

Can we add the example template, to the templates folder in this repo?

2.0/runtime/README.md
tmds: > If you change the environment variable in the runtime image that change should carry over to any subsequent images. So a runtime image built with the HTTP_PROXY setting configured should still have that configuration for any application that depends on the runtime image.

That is true for any environment variable.

> It's also the only setting in the build image (other than ASPNETCORE_URLS) that still has some relevance outside of build time for that reason, hence why I kept it here.

If the multi-image build doesn't carry over these variables (TBD), we should add a note (as part of the openshift multi-image instructions) on how we recommend handling this. This is specific on how the multi-image build differs from the single-image build.

jerboaa:
[cibuild]

jerboaa:
If there are no objections I'll merge this. It's already huge. We can fix smaller things in subsequent patches. @aslicerh @omajid @tmds Thoughts?

tmds:
ok to merge

tmds:
I hope there is some doc on how to use this in OpenShift soon as I want to try it and see how this looks to the end-user.

aslicerh:
Let me make the readme change @tmds wanted and then we'll merge.

jerboaa:
Thanks @aslicerh!

PR #90, opened 2017-06-29
Support runtime images with stores, disable ASP.NET Core store by default

tmds:
Implements https://github.com/redhat-developer/s2i-dotnetcore/issues/86


jerboaa:
Looks OK to me.

tmds:
> setting this option to true will break things.

Not if the user has a runtime image with an asp.net core store.

It also makes it possible to use a custom store, this was inhibited by the previous implementation since we forced all manifests to be ignored.

PR #94, opened 2017-07-10
Allow tests to be run stand-alone.

jerboaa:
This adjusts scripts so that tests can either be run
directly in the test directory of the respective image
or by calling the test/run script from any directory by
using the path to the test/run script.

The latter wasn't possible prior this patch.


jerboaa:
Thanks for the review!

Issue #95, opened 2017-07-14
runtime image does not have an entrypoint


tmds:
To use the runtime image, you can do this:
```
FROM registry.access.redhat.com/dotnet20-runtime-rhel7

ADD bin/Release/netcoreapp2.0/rhel.7-x64/publish/. .

CMD ["scl", "enable", "rh-dotnet20", "--", "dotnet", "couchbase_demo.dll"]
```
It would be nicer without the scl enable.
We can probably handle this via an ENTRYPOINT.
This should also call generate_container_user.

The entrypoint could also check for the s2i_run script and invoke that when there are no arguments. Then the user doesn't need to handle this for a s2i build into a runtime image.

tmds:
Andrew, can you look at this? I will be on pto for the next weeks.
CC @jerboaa @omajid

jerboaa:
Thanks for filing the bug @tmds

jerboaa:
The reason why the "scl enable" is needed because there is no defined entrypoint which goes through bash. For example `CMD ['bash', '-c', 'dotnet couchbase_demo.dll'] works fine. The builder has an [explicit entrypoint](https://github.com/redhat-developer/s2i-dotnetcore/blob/2.0-dev/2.0/build/Dockerfile.rhel7#L62) that goes through bash. Perhaps we should move those bits from the builder image to the runtime image.

@aslicerh Thoughts?

aslicerh:
@jerboaa That sounds like a good idea. I'm at a good stopping point in what I'm working on, so I'll have a PR for this up next.

jerboaa:
Fixed with PR #96.

tmds:
@jerboaa @aslicerh the PR doesn't include

> This should also call generate_container_user.

Any thoughts on including this in the entrypoint script?

aslicerh:
@tmds I was going to discuss that with you when you came back. The generate_container_user script adds an entry into the nss passwd file every time it is run. If you add it to the ENTRYPOINT, then each time the same container is run it will keep adding more and more duplicate entries.

We should either find a way to add it to the assembly of the container, or we need to change/create the script to check if the entry already exists and add it if it doesn't.

PR #96, opened 2017-07-26
Move ENTRYPOINT definition to the runtime image and add a test for it.

aslicerh:
This should address issue #95.


2.0/build/Dockerfile.rhel7
```diff
@@ -59,7 +59,5 @@ WORKDIR /opt/app-root/src
 # Run container by default as user with id 1001 (default)
 USER 1001
 
-ENTRYPOINT ["container-entrypoint"]
```
jerboaa: Could we also move the `COPY ./root/usr/bin /usr/bin` to the runtime image? Those are all helpers useful in the runtime image too and are rather low in size. It would reduce the build image instructions as well.

2.0/runtime/root/usr/bin/container-entrypoint
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
jerboaa: Could we mark this as a rename perhaps? Not sure if this is a rename already. `git mv 2.0/build/root 2.0/runtime/`

2.0/build/Dockerfile.rhel7
aslicerh: The usage script doesn't make sense to move over, because it is about s2i usage, but the s2i usage doesn't make sense because you can't use s2i to build packages on the runtime.

2.0/runtime/root/usr/bin/container-entrypoint
aslicerh: If you look at the actual diff in git, it already says it is a 'rename'.

jerboaa: OK, cool. Sorry for the noise.

2.0/build/Dockerfile.rhel7
jerboaa: Are you referring to `base-usage` script? I don't think we need this, do we? There is [usage](https://github.com/redhat-developer/s2i-dotnetcore/blob/2.0-dev/2.0/build/s2i/bin/usage) which is in the `s2i` folder. Perhaps we can remove the `base-usage` script entirely and move everything else to runtime?

2.0/runtime/root/usr/bin/container-entrypoint
aslicerh: No worries.

2.0/build/Dockerfile.rhel7
aslicerh: Yeah, I don't see it being used anywhere, and it didn't even exist in the older versions. Ok, I'll get rid of that one and move the other two.

aslicerh: Done.

jerboaa:
Looks good to me. I'll merge it in.

PR #97, opened 2017-08-07
Build and test with latest internal 2.0 build

tmds:
* provide repos as part of `yum install`, using `yum-config-manager` the repo doesn't always get enabled on my machine 
* Hard code myget feeds to make the tests work
* Add instruction to add repo file for internal build (repo file not included)
* runtime/test: minimal testing of dotnet, remove pre-compiled tar.gz applications


jerboaa:
@tmds A couple of things come to mind: Why remove pre-compiled examples for runtime tests? To me it makes sense to have them. We have the readme which describes how to generate them.

Shouldn't we separate changes needed to a) work with latest dotnet 2.0 version b) to get actual repos set up. It seems it would be OK to require a local patch to build with internal repos. Thoughts?

tmds:
> @tmds A couple of things come to mind: Why remove pre-compiled examples for runtime tests? To me it makes sense to have them. We have the readme which describes how to generate them.

If we want these tests, the generation should be part of the test. I don't like the pre-compiled dotnet applications packed as tar.gz.

> Shouldn't we separate changes needed to a) work with latest dotnet 2.0 version b) to get actual repos set up. It seems it would be OK to require a local patch to build with internal repos. Thoughts?

The separation is nice, if we are all applying the same patches locally, we may put it in the repo. I can remove the COPY rh-dotnet20.repo if you like.

jerboaa:
> If we want these tests, the generation should be part of the test. I don't like the pre-compiled dotnet applications packed as tar.gz.

It's a fair point, but we want to be able to test runtime and builder image independently. The builder image is all that can be assumed to exist. That also means there is no compiler as far as I'm aware (no sdk). Tests would mimic actual use of the image, which it does by running those pre-compiled apps.

> The separation is nice, if we are all applying the same patches locally, we may put it in the repo. I can remove the COPY rh-dotnet20.repo if you like.

Let's first get an update in for the first problem (make it work with latest bits). Then discuss the repo changes to do builds separately.

tmds:
> It's a fair point, but we want to be able to test runtime and builder image independently. The builder image is all that can be assumed to exist. That also means there is no compiler as far as I'm aware (no sdk). Tests would mimic actual use of the image, which it does by running those pre-compiled apps.

It is testing something that resembles the OpenShift split build where the artifact comes from the sdk image and is fed as a tar.gz file to the runtime image. So it would be good to generate the tar.gz files of the sdk image.

tmds:
> I'm OK with changes to assemble and build/test/run, but other changes are somewhat separate concerns.

Do you want me to undo the changes to the runtime test?

jerboaa:
> Do you want me to undo the changes to the runtime test?

Yes, please.

jerboaa:
When you merge, please squash appropriately.

tmds:
> Looks good to me. Could you please create a ticket for removing myget feeds?

There are still some other TODOs which don't have tickets.

> When you merge, please squash appropriately.

I want to resolve the nss_wrapper issue before merging this. See mail I sent. With the `yum-config-manager` syntax, the package doesn't install in the image on (what I believe to be) a properly subscribed machine.

tmds:
@jerboaa I removed the tests again, since they are actually failing with the updated 2.0 rpm:
```
It was not possible to find any compatible framework version
The specified framework 'Microsoft.NETCore.App', version '2.0.0-preview1-002111-00' was not found.
  - Check application dependencies and target a framework version installed at:
      /
  - Alternatively, install the framework version '2.0.0-preview1-002111-00'.
```

jerboaa:
@tmds It would be better to regenerate the tests with updated framework, since we'll need them to verify the runtime image works anyway.

tmds:
> @tmds It would be better to regenerate the tests with updated framework, since we'll need them to verify the runtime image works anyway.

We need a test for the OpenShift use-case. And we need a test for the runtime image (+ script to generate binary). I prefer to do this separate from this PR.

tmds:
Changes to the yum install are a workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1479388.

jerboaa:
Feel free to merge.

2.0/runtime/test/run
```diff
@@ -6,426 +6,31 @@
...
-  if [ -n ${TMPDIR} ] && [ -e ${TMPDIR} ]; then
-    echo "Removing temp directory ${TMPDIR}"
-    rm -rf ${TMPDIR}
-  fi
-}
-
-# This tests that the container has an ENTRYPOINT defined such that
-# a new image built FROM the runtime image can have a CMD that simply
-# calls dotnet to run an app directly.
-test_entrypoint() {
```
jerboaa: Ugh, that test got removed too :( @tmds Please re-enable it. It's a regression test for #95.

2.0/runtime/test/run
```diff
@@ -434,34 +39,7 @@ info "Testing ${IMAGE_NAME}"
 # it from Docker hub
 s2i_args="--force-pull=false"
 
-test_entrypoint
```
jerboaa: Same, that was a useful test.

PR #98, opened 2017-08-08
Refactor responsibilities runtime & build image

tmds:
* replace DOTNET_PUBLISH_PATH and DOTNET_RUN_SCRIPT of builder with DOTNET_APP_PATH and DOTNET_DEFAULT_CMD in runtime.
DOTNET_APP_PATH is at app-root/app and is the WORKDIR for the runtime image.
DOTNET_DEFAULT_CMD is default-cmd.sh and has a default implementation which echos some info
* add user in runtime instead of build
* call generate_container_user in runtime entrypoint when PID == 1
note: generate_container_user creates a new user file for each invocation, it doesn't append
* Update runtime README example to show how to create a Dockerfile from a published app output.
* runtime/test: test nss_wrapper and default-cmd.sh

### DIFF
--- a/2.0/runtime/README.md
+++ b/2.0/runtime/README.md
@@ -33,13 +33,10 @@ $ docker build -t s2i-dotnetcore-ex .
 
 Start a container:
 ```
-$ docker run -p 8080:8080 s2i-dotnetcore-ex
+$ docker run --rm -p 8080:8080 s2i-dotnetcore-ex
 ```
 
-Fetch a webpage:
-```
-$ curl http://127.0.0.1:8080
-```
+Visit the web application that is running in the docker container with a browser at [http://localhost:8080].
 
 Repository organization
 ------------------------
--- a/2.0/runtime/README.md
+++ b/2.0/runtime/README.md
@@ -9,38 +9,34 @@ The image can be run using [Docker](http://docker.io).
 Usage
 ---------------------
 The distributale binaries of an application should be copied to this image and
-a new docker image should be created using this image as the base. As an
-example, given an archive of the binaries of a simple [dotnet-sample-app](test/asp-net-hello-world)
-application, the following Dockerfile could be used.
+a new docker image should be created using this image as the base.
 
-**Sample Dockerfile**
+For example to create a Docker image for [s2i-dotnetcore-ex](https://github.com/redhat-developer/s2i-dotnetcore-ex) 
+
+Publish the application:
 ```
-FROM dotnet-20-rhel7-runtime
-RUN mkdir -p '/opt/app-root/publish'
-WORKDIR /opt/app-root/publish
-COPY asp-net-hello-world.tar.gz asp-net-hello-world.tar.gz
-RUN tar xvf asp-net-hello-world.tar.gz
-CMD /opt/app-root/publish/s2i_run
+$ git clone -b dotnetcore-2.0 https://github.com/redhat-developer/s2i-dotnetcore-ex.git
+$ cd s2i-dotnetcore-ex/app
+$ dotnet restore -r rhel.7-x64
+$ dotnet publish -f netcoreapp2.0 -c Release -r rhel.7-x64 --self-contained false /p:PublishWithAspNetCoreTargetManifest=false
 ```
 
-As for generating the binary archives in the first place, often this can be
-accomplished by running a dotnet restore and a dotnet publish to a single
-directory and then archiving that directories contents.
-
-**Example Binary Archive Creation**
-``` sh
-cd asp-net-hello-world
-dotnet restore "." -r "rhel.7-x64"
-dotnet publish "." -f "netcoreapp2.0" -c "Release" -r "rhel.7-x64" --self-contained false /p:TargetManifestFiles= -o "./publish"
-tar zcvpf "./asp-net-hello-world.tar.gz" -C ./publish .
+Create the Docker image:
+```
+$ cat > Dockerfile <<EOF
+FROM dotnet/dotnet-20-runtime-rhel7
+ADD bin/Release/netcoreapp2.0/rhel.7-x64/publish/. .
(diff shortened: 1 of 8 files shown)


2.0/runtime/README.md
```diff
@@ -9,38 +9,34 @@ The image can be run using [Docker](http://docker.io).
...
 ```
 
-This example cds into the project's directory, restores the projects NuGet dependencies, and then publishes the project to a publish directory. Finally, a tar command is used to create a gzipped tar file containing the content of the publish directory. This tar file is what is copied and extracted to create the apps runtime image, using the sample dockerfile given above.
-
-**Accessing the application:**
-
-HTTP:
+Start a container:
+```
+$ docker run -p 8080:8080 s2i-dotnetcore-ex
```
jerboaa: This should probably be: `docker run -p 8080:8080 -d s2i-dotnetcore-ex`. Run as daemon.

2.0/runtime/README.md
```diff
@@ -9,38 +9,34 @@ The image can be run using [Docker](http://docker.io).
...
+```
 
+Fetch a webpage:
```
jerboaa: I'd suggest to use as last step:

    View the running application within the docker container with a browser at: http://127.0.0.1:8080

2.0/runtime/default-cmd.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+cat <<EOF
+This is a runtime image for .NET Core ${DOTNET_CORE_VERSION}.
+
+See the documentation at https://github.com/redhat-developer/s2i-dotnetcore.
```
jerboaa: This should probably link to the runtime image docs directly:

      See the documentation at https://github.com/redhat-developer/s2i-dotnetcore/tree/master/2.0/runtime

2.0/runtime/root/usr/bin/container-entrypoint
```diff
@@ -1,12 +1,10 @@
...
-if [ $# -eq 0 ]; then
-	# No arguments given. Try to use s2i_run, if it exists.
-	if [ -f /opt/app-root/publish/s2i_run ]; then
-		exec /opt/app-root/publish/s2i_run
-	fi
-else
-	cmd="$1"; shift
-	exec $cmd "$@"
+
+if [ $$ -eq 1 ]; then
```
jerboaa: If PID == 1 source generate_container_user. Why on this condition? It should be PID==1 in 99.9% of the time this is invoked.

I'd rather source it unconditionally, and check in the script (`generate_container_user`) whether an entry in `/opt/app-root/etc/passwd` for the current user ID exists. If not, generate one, otherwise no-op.

tmds: This is to not generate_container_user for exec into a running container.

2.0/runtime/default-cmd.sh
tmds: It should be easy to navigate to the right page. I don't want to fix the location of the runtime README on our master branch.

jerboaa: I disagree. It might be easy for us to find, not so for our users. There is a chance for users to navigate to the wrong version. Once released I don't expect the runtime README to change location. So we might as well point them there. Why is fixing to master a concern? If there is 2.1, then there will be a 2.1 folder. 2.0 will still exist.

2.0/runtime/README.md
tmds: Going to add --rm here so the container gets removed. Will update the last step to using browser. So it's ok to not run as daemon here: makes it easy to stop and clean up the container.

2.0/runtime/root/usr/bin/container-entrypoint
jerboaa: If the generate_container_user script gets modified to not generate an entry for the current UID if it already exists, then this should not matter. It will have generated one when the app starts and the exec will not generate a new one.

tmds: The current implementation has the advantage it avoids race conditions due to multiple processes checking and updating the users file.

2.0/runtime/README.md
jerboaa: Works for me.

2.0/runtime/root/usr/bin/container-entrypoint
jerboaa: I'm doubtful this is a problem in practice.

2.0/runtime/default-cmd.sh
tmds: There is a difference between not expecting it to change and disallowing it to ever change.
The readme should be user friendly enough to find the proper documentation. The message contains the version of the runtime image so that will help find the proper version.
If you feel strongly about it, I'll update the url.

2.0/runtime/root/usr/bin/container-entrypoint
tmds: Unless we expect an issue with the current implementation, I prefer to keep it because it's simpler and more robust.

2.0/runtime/default-cmd.sh
jerboaa: The [imagestreams](https://github.com/redhat-developer/s2i-dotnetcore/blob/master/dotnet_imagestreams.json#L26) have references to master as well. Providing a direct link is more user friendly. Yes, please update it.

2.0/runtime/root/usr/bin/container-entrypoint
jerboaa: If you want to keep it, please add a comment that you are checking for PID==1 because of the exec'ing into the container use-case.

tmds:
@jerboaa comments are addressed, can you approve the PR if ok to merge?

tmds:
Thanks!

PR #100, opened 2017-08-08
Auto-generate runtime image test apps.

aslicerh:
This also restores the ENTRYPOINT test that had been removed.

### DIFF
--- a/2.0/gen_tests.sh
+++ b/2.0/gen_tests.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+#
+# Generate the test tarballs for the runtime image
+#
+# PRE: Docker daemon running, s2i binary available,
+#      Build image available
+#
+# Supported environment variables:
+#
+# -----------------------------------------------------
+#
+#  TEST_SRC_DIR     -> Where the test app source directories are.
+#                      Defaults to ./build/test
+#  TEST_TAR_DIR     -> Where the generated tarballs should go.
+#                      Defaults to ./runtime/test
+#  TEST_LIST        -> The list of apps to try and build tarballs for
+#                      See the script for the defaults
+#  BUILD_IMAGE      -> The tag of the build image to use to build the apps.
+#                      Defaults to dotnet/dotnet-20-rhel7
+#  TEST_COMMIT_HASH -> The commit hash to use to use to check if apps need to
+#                      be rebuilt (and relinked).
+#                      Defaults to the hash of the currently defined HEAD.
+# -----------------------------------------------------
+#
+
+if [ "$DEBUG" != "" ]; then
+  set -x
+fi
+
+TEST_SRC_DIR=${TEST_SRC_DIR:-`dirname $(readlink -m $0)`/build/test}
+TEST_TAR_DIR=${TEST_TAR_DIR:-`dirname $(readlink -m $0)`/runtime/test}
+TEST_LIST=${TEST_LIST:-hw_framework_config helloworld asp-net-hello-world}
+BUILD_IMAGE=${BUILD_IMAGE:-dotnet/dotnet-20-rhel7}
+TEST_COMMIT_HASH=${TEST_COMMIT_HASH:-`git rev-parse HEAD`}
+
+info() {
+  echo -e "\n\e[1m[INFO] $@\e[0m\n"
+}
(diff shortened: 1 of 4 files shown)


aslicerh:
Alright, here is a first attempt at auto-generating the test tarballs.

I'm still on the fence if I like the hardlink stuff I did to normalize the tar names without getting rid of the tarballs.

Part of the reason that I want normalized tarball names is because the rh-dotnetcore containers will need the tarballs for tests, but they will have a different commit hash than the equivalent s2i-dotnetcore commit.

2.0/gen_tests.sh
```diff
@@ -0,0 +1,80 @@
...
+  app_tar="${app}-${TEST_COMMIT_HASH}.tar.gz"
+
+  if [ -f ${TEST_TAR_DIR}/${app_tar} ]; then
+    info "Tar ${app_tar} already exists. Skipping."
+    continue
+  fi
+
+  info "Generating ${app_tar}"
+
+  tar_dir_mnt_path="/opt/tar"
```
jerboaa: Doesn't this assume `/opt/{tar,src}` exists? Could we perhaps use temporary directories for those? I'm thinking `mktemp -d`. Then move the tar to the expected location in the `test` folder and finally create a file with containing the commit SHA so as to be able to recreate the sources.

2.0/gen_tests.sh
```diff
@@ -0,0 +1,80 @@
...
+  src_dir_mnt_path="/opt/src"
+  test_image="${BUILD_IMAGE}-${app}"
+  tar_cmd="tar -C /opt/app-root/publish -cvf ${tar_dir_mnt_path}/${app_tar} ."
+
+  s2i build --force-pull=false "file://${src_path}" ${BUILD_IMAGE} ${test_image}
+  docker run -v=${src_path}:${src_dir_mnt_path} -v=${TEST_TAR_DIR}:${tar_dir_mnt_path} --rm ${test_image} ${tar_cmd}
+  docker rmi ${test_image}
+done
+
+# Create hard links to the current tests
```
jerboaa: I'm not sure we need this at all. Perhaps we can also generate the sources as `{app}-src.tar.gz`, which live side-by-side in the test directory. The tests would only use the compiled versions, if things go wrong one can look at the source tarball. Belts and braces solution would be to also record the commit sha in a file so as to be able to regenerate the sources.

tmds:
I was thinking about having a simple test in the runtime image (instead of the tar.gz stuff) and generate on the fly some tar.gz files to test the OpenShift case in the sdk test.
This is what I was saying on irc yesterday, but the PR here is doing something different.

tmds:
something like:
```
runtime/test/hello-world/hello-world.dll <- simple app to test the runtime image (even not packed as tar.gz)
sdk/test/run -> for some of the apps, generate tar.gz file and feed that to the runtime same way as OpenShift split use-case
```

2.0/gen_tests.sh
aslicerh: They don't have to exist ahead of time. They will be created in the docker container since they are used as part of the host volume mount option.

The point of these is to give the container access to the host directories for the source and a place (on the host) to dump the resulting tarball.

2.0/gen_tests.sh
aslicerh: Just to explain this weirdness, let me tell you what I was thinking.

So the idea for a good chunk of this was a way to get tests without checking the tarballs into source, but also without regenerating them on every build. So the tarballs are marked with the commit hash they were built from and only rebuilt if the current hash changes.

But what if you want to test against a different commit? Like if a test is failing now and you want to make sure it wasn't before? Well, just building the tarballs doesn't necessarily help and is something of a pain. Plus, the commit hashes in this repo will not match the hashes for the rh-dotnetcore container build repos.

So my thought was that using hardlinks like this would mean that the tests could switch between hashes without deleting and rebuilding, and it meant that the rh-dotnetcore containers would use the same test script without having to know about hashes that weren't related to that repo.

None of this may matter now, but I still wanted to explain what I was trying to get past with something odd like this.

jerboaa:
I'm closing this one. Please reopen if this is still something we'd want merged.

PR #101, opened 2017-08-09
Refactor tests and add ASP.NET Core test

tmds:
- Refactor tests by grouping them into test functions and calling assert functions to check
- Add test for default command, port and version
- Add test for running ASP.NET Core app on runtime image
The test is pre-compiled in the bin folder. The build-project.sh script can be invoked to build the test project.


tmds:
The important file to review for this pr is the test and build-project.sh scripts. The other files make up the aspnet-hello-world project and can be ignored.

tmds:
For the test script, it may be simpler to review the first commit: https://github.com/redhat-developer/s2i-dotnetcore/pull/101/commits/747f550107351534dcea217aaca0cb507906aa00
and then the changes made for the aspnet test: https://github.com/redhat-developer/s2i-dotnetcore/pull/101/commits/fcc0e40fc60811d96f4b4f72cba7a769f52952a2#diff-44fcafd39c2e1f5aeec48c191e51ca4f.

tmds:
I've updated the commit to pack the output in a tar.gz file, it's a lot easier to review now.

2.0/runtime/test/run
```diff
@@ -21,45 +23,141 @@ info() {
...
+  test_start
+
+  pushd ${test_dir}/aspnet-hello-world
+
+  # create image
+  local container=dotnet-helloworld
+  local image=${container}
+  cat >Dockerfile <<EOF
+FROM ${IMAGE_NAME}
+ADD app.tar.gz .
```
jerboaa: I don't quite get how this is going to work. Where is this being extracted? Is there some magic being involved?

tmds: It's a docker feature. If you ADD a tar.gz file, it will auto-magically extract it.

jerboaa: Hah! I see, thanks.

aslicerh: It is generally recommended to not use ADD, because of how easily it changes behaviour based on context:
https://github.com/moby/moby/issues/3050

PR #102, opened 2017-08-09
Added runtime openshift template and documentation.

aslicerh:


aslicerh:
Note: This won't actually work until the 2.0 containers are available. But you can grab a local copy and change the container it points to test it until then.

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+                }
+            },
+            "spec": {
+                "source": {
+                    "dockerfile": "FROM ${NAME}:2.0-runtime \nCOPY publish/ /opt/app-root/publish\n\nWORKDIR /opt/app-root/publish\nCMD \"/opt/app-root/publish/s2i_run\"",
+                    "images": [
+						{
+							"from": {
+								"kind": "ImageStreamTag",
+								"name": "${APP_BUILD_IMAGE_TAG}"
```
tmds: Can we include this BuildConfig too, this is the one that uses the dotnet build image.

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+                "tags": [
+                    {
+                        "name": "2.0-runtime",
+                        "annotations": {
+                            "openshift.io/display-name": ".NET Core 2.0 Runtime",
+                            "description": "Run .NET Core 2.0 applications on RHEL 7. For more information about using this runtime image, including OpenShift considerations, see https://github.com/redhat-developer/s2i-dotnetcore/tree/master/2.0/README.md.",
+                            "iconClass": "icon-dotnet",
+                            "tags": "runtime,.net,dotnet,rh-dotnet20",
+                            "supports":"dotnet:2.0-runtime,dotnet",
+                            "sampleRepo": "https://github.com/redhat-developer/s2i-dotnetcore-ex.git",
```
jerboaa: Those `sample*` fields can be removed. I don't think they'll have any effect.

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+            "apiVersion": "v1",
+            "metadata": {
+                "name": "${NAME}",
+                "annotations": {
+                    "description": "Defines how to chain the runtime image from the source build image"
+                }
+            },
+            "spec": {
+                "source": {
+                    "dockerfile": "FROM ${NAME}:2.0-runtime \nCOPY publish/ /opt/app-root/publish\n\nWORKDIR /opt/app-root/publish\nCMD \"/opt/app-root/publish/s2i_run\"",
```
tmds: This should be cleaned up. There is no more app-root/publish folder and no s2i_run script.

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+                    "images": [
+						{
+							"from": {
+								"kind": "ImageStreamTag",
+								"name": "${APP_BUILD_IMAGE_TAG}"
+							},
+							"paths": [
+								{
+									"sourcePath": "/opt/app-root/publish",
```
tmds: "sourcePath": "/opt/app-root/app"

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+                    "description": "Keeps track of changes in the application image"
+                }
+            },
+            "spec": {
+                "tags": [
+                    {
+                        "name": "2.0-runtime",
+                        "annotations": {
+                            "openshift.io/display-name": ".NET Core 2.0 Runtime",
+                            "description": "Run .NET Core 2.0 applications on RHEL 7. For more information about using this runtime image, including OpenShift considerations, see https://github.com/redhat-developer/s2i-dotnetcore/tree/master/2.0/README.md.",
```
jerboaa: `https://github.com/redhat-developer/s2i-dotnetcore/tree/master/2.0/README.md` does not exist. Not even in the 2.0-dev branch. There are only READMEs for build/runtime. Perhaps link to `2.0/runtime/README.md`?

README.md
```diff
@@ -76,6 +76,14 @@ have completed, you can browse to the application using the url you find in proj
 The dotnet-example template can be used to create a new .NET Core service in OpenShift. It provides parameters for all the environment
 variables of the s2i-dotnetcore builder. It also includes a liveness and a readiness probe.
 
+**dotnet-runtime-example**
+
+The dotnet-runtime-example template can be used to create a new .NET Core service in OpenShift. It is meant to serve as an example of
+how to create a 'chained build' where one image is used to build an application but the runtime image is used to deploy the application.
+This service will deploy the application built by the dotnet-example template above, so that must also be created.
+
+For more information on build chaining in OpenShift, [please see this doc](ihttps://docs.openshift.org/latest/dev_guide/builds/advanced_build_operations.html#dev-guide-chaining-builds).
```
jerboaa: Typo: `ihttps` => `https`

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+                                },
+                                "env": []
+                            }
+                        ]
+                    }
+                }
+            }
+        }
+    ],
+    "parameters": [
```
tmds: I'd be nice if this was the same parameters as the dotnet-examples. No parameters added/removed. Just the way the image is built should be different.

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+            "metadata": {
+                "name": "${NAME}",
+                "annotations": {
+                    "description": "Keeps track of changes in the application image"
+                }
+            },
+            "spec": {
+                "tags": [
+                    {
+                        "name": "2.0-runtime",
```
tmds: This is not a good name. "dotnet-runtime".
Why don't we provide this as part of our imagestreams?

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+                        "annotations": {
+                            "openshift.io/display-name": ".NET Core 2.0 Runtime",
+                            "description": "Run .NET Core 2.0 applications on RHEL 7. For more information about using this runtime image, including OpenShift considerations, see https://github.com/redhat-developer/s2i-dotnetcore/tree/master/2.0/README.md.",
+                            "iconClass": "icon-dotnet",
+                            "tags": "runtime,.net,dotnet,rh-dotnet20",
+                            "supports":"dotnet:2.0-runtime,dotnet",
```
tmds: "supports":"dotnet-runtime:2.0,dotnet:2.0,dotnet"

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+                            "sampleRepo": "https://github.com/redhat-developer/s2i-dotnetcore-ex.git",
+                            "sampleContextDir": "app",
+                            "sampleRef": "dotnet-2.0-runtime",
```
tmds: "dotnetcore-2.0"

templates/dotnet-runtime-example.json
jerboaa: It was a suggestion from myself. This image stream is not needed for any other usecase than this template. The reasoning was that it would have a dep on our dotnet imagestreams being present. I'm fine with moving it there if that's the consensus.

templates/dotnet-runtime-example.json
tmds: For the buildconfigs, use these names: ${NAME}-build and ${NAME}-runtime

templates/dotnet-runtime-example.json
jerboaa: There is no real way to "try" this imagestream in OpenShift other than with this template, hence my suggestion to remove those sample* items entirely. Contrast this with the builder image which can build and run an app stand-alone.

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+							]
+						}
+					]
+                },
+                "strategy": {
+                    "type": "Docker",
+                    "dockerStrategy": {
+                        "from": {
+                            "kind": "ImageStreamTag",
+                            "name": "${NAME}:2.0-runtime"
```
tmds: Shouldn't this refer to the dotnet runtime image? "2.0-runtime:2.0" (-> suggested rename: "dotnet-runtime:2.0")

templates/dotnet-runtime-example.json
tmds: FROM instruction should refer to the dotnet runtime. "2.0-runtime:2.0" (-> suggested rename: "dotnet-runtime:2.0")

templates/dotnet-runtime-example.json
tmds: I prefer to move this to the image streams because it's generic and re-usable across templates. Also we can provide proper naming and metadata.

templates/dotnet-runtime-example.json
tmds: +1

templates/dotnet-runtime-example.json
tmds: You can ignore this, I was confused by the naming.

templates/dotnet-runtime-example.json
tmds: You can the comment for the FROM instruction, I was confused by the naming.

templates/dotnet-runtime-example.json
jerboaa: OK. I'll add it to #99. Andrew, feel free to remove it from this one.

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+        },
+        {
+            "name": "APP_BUILD_IMAGE_TAG",
+            "displayName": "Application build image",
+            "description": "The image stream tag which is used to build the application.",
+            "required": true,
+            "value": "dotnet-example:latest"
+        },
+        {
+            "name": "DOTNET_IMAGE_STREAM_RUNTIME_TAG",
```
jerboaa: While trying this out, I"ve noticed this isn't used anywhere in the template. Could you fix this, please? Suggested param name: `DOTNET_RUNTIME_IMAGE_STREAM_TAG`

tmds: @jerboaa I suggested this template to have exactly the same parameters as the dotnet-example and only differ in the way it is built.

tmds: ah, we do need an extra parameter for the runtime. the DOTNET_IMAGE_STREAM_TAG is used in the dotnet-example for the build image. DOTNET_RUNTIME_IMAGE_STREAM_TAG seems is a proper name :+1:

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+            "displayName": ".NET Runtime",
+            "description": "The image stream tag which is used to run the application.",
+            "required": true,
+            "value": "dotnet-runtime-example:2.0-runtime"
+        },
+        {
+            "name": "NAMESPACE",
```
jerboaa: Same with this one. It's not used anywhere. It should be used for the recources not necessarily coming from the `openshift` namespace. In this case the DOTNET_RUNTIME_IMAGE_STREAM_TAG

templates/dotnet-runtime-example.json
```diff
@@ -0,0 +1,263 @@
...
+                "name": "${NAME}",
+                "annotations": {
+                    "description": "Defines how to chain the runtime image from the source build image"
+                }
+            },
+            "spec": {
+                "source": {
+                    "dockerfile": "FROM ${NAME}:2.0-runtime \nCOPY publish/ /opt/app-root/publish\n\nWORKDIR /opt/app-root/publish\nCMD \"/opt/app-root/publish/s2i_run\"",
+                    "images": [
+						{
```
aslicerh: Sorry about that. I need to change my vim config over to just always using spaces at this point, I think.

templates/dotnet-example.json
```diff
@@ -269,7 +269,7 @@
             "name": "SOURCE_REPOSITORY_REF",
             "displayName": "Git Reference",
             "description": "Set this to a branch name, tag or other ref of your repository if you are not using the default branch.",
-            "value": "dotnetcore-1.0"
+            "value": "dotnetcore-2.0"
```
aslicerh: Yeah, I just noticed that too. I'll remove that.

templates/dotnet-runtime-example.json
aslicerh: Ah. This line is copied from the dotnet-example.json file, so it probably needs to be changed there as well.

templates/dotnet-runtime-example.json
aslicerh: Why do you say there is no app-root/publish or s2i_run script?

README.md
aslicerh: Done.

templates/dotnet-runtime-example.json
aslicerh: Ok, I'll remove it from here.

templates/dotnet-runtime-example.json
jerboaa: https://github.com/redhat-developer/s2i-dotnetcore/commit/b52a3fd2b267c274b33742bfd36d02c14e09c58c

jerboaa:
Andrew, please make this a PR for master (or create a new one). We are going to merge 2.0-dev into master any time now.

aslicerh:
@jerboaa If you go ahead and do the merge I'll switch the PR over. I don't want to switch it before then.

jerboaa:
@aslicerh sounds good.

aslicerh:
Ok, I think this will have addressed everything so far.

Because of the changes to s2i, this one won't even work with the dotnet-beta 2.0 containers without some more edits to a local copy.

templates/dotnet-runtime-example.json
```diff
@@ -56,27 +56,101 @@
...
+                            {
+                                "name": "DOTNET_TEST_PROJECTS",
+                                "value": "${DOTNET_TEST_PROJECTS}"
+                            },
+                            {
+                                "name": "DOTNET_CONFIGURATION",
+                                "value": "${DOTNET_CONFIGURATION}"
+                            },
+                            {
+                                "name": "DOTNET_PUBLISH",
```
jerboaa: For this new template we can remove this env variable. We are always publishing in 2.0+

templates/dotnet-runtime-example.json
```diff
@@ -86,35 +160,36 @@
...
-                "name": "${NAME}",
+                "name": "${NAME}-runtime",
                 "annotations": {
                     "description": "Defines how to chain the runtime image from the source build image"
                 }
             },
             "spec": {
                 "source": {
-                    "dockerfile": "FROM ${NAME}:2.0-runtime \nCOPY publish/ /opt/app-root/publish\n\nWORKDIR /opt/app-root/publish\nCMD \"/opt/app-root/publish/s2i_run\"",
+                    "dockerfile": "FROM ${DOTNET_RUNTIME_IMAGE_STREAM_TAG}\nCOPY publish/ /opt/app-root/app\nWORKDIR /opt/app-root/app\nUSER 0\nRUN chown -R 1001:0 /opt/app-root/app && chmod -R og+rwx /opt/app-root/app\nUser 1001",
```
jerboaa: `RUN fix-permissions /opt/app-root/app` should work with latest runtime image as it'll have the `fix-permissions` script.

README.md
```diff
@@ -76,6 +76,14 @@ have completed, you can browse to the application using the url you find in proj
 The dotnet-example template can be used to create a new .NET Core service in OpenShift. It provides parameters for all the environment
 variables of the s2i-dotnetcore builder. It also includes a liveness and a readiness probe.
 
+**dotnet-runtime-example**
+
+The dotnet-runtime-example template can be used to create a new .NET Core service in OpenShift. It is meant to serve as an example of
+how to create a 'chained build' where one image is used to build an application but the runtime image is used to deploy the application.
+This service will deploy the application built by the dotnet-example template above, so that must also be created.
```
jerboaa: This comment is no longer correct. You may refer to it building and deploying `redhat-developer/s2i-dotnetcore-ex`

templates/dotnet-example.json
```diff
@@ -4,8 +4,8 @@
     "metadata": {
         "name": "dotnet-example",
         "annotations": {
-            "openshift.io/display-name": ".NET Core",
-            "description": "An example .NET Core application.",
+            "openshift.io/display-name": ".NET Core Example",
```
jerboaa: Please remove those changes. We can add them in PR #103.

templates/dotnet-runtime-example.json
tmds: or:
"FROM ${DOTNET_RUNTIME_IMAGE_STREAM_TAG}\nADD app.tar.gz ."
with DOTNET_PACK=true

jerboaa: Right. Would mimic what we are testing.

templates/dotnet-runtime-example.json
```diff
@@ -167,7 +167,7 @@
             },
             "spec": {
                 "source": {
-                    "dockerfile": "FROM ${DOTNET_RUNTIME_IMAGE_STREAM_TAG}\nCOPY publish/ /opt/app-root/app\nWORKDIR /opt/app-root/app\nUSER 0\nRUN chown -R 1001:0 /opt/app-root/app && chmod -R og+rwx /opt/app-root/app\nUser 1001",
+                    "dockerfile": "FROM ${DOTNET_RUNTIME_IMAGE_STREAM_TAG}\nWORKDIR /opt/app-root/app\nADD app.tar.gz\nUser 1001",
```
tmds: Why do we need a "User" statement?

tmds: Or a "WORKDIR" statement?

tmds: The sourcePath below needs to be adjusted "/opt/app-root/app.tar.gz" too

aslicerh: Both the user and workdir statements are leftovers from things that are no longer true. I'll remove them.

And I went ahead and changed the sourcePath.

tmds: :+1:

templates/dotnet-runtime-example.json
```diff
@@ -167,7 +167,7 @@
             },
             "spec": {
                 "source": {
-                    "dockerfile": "FROM ${DOTNET_RUNTIME_IMAGE_STREAM_TAG}\nWORKDIR /opt/app-root/app\nADD app.tar.gz\nUser 1001",
+                    "dockerfile": "FROM ${DOTNET_RUNTIME_IMAGE_STREAM_TAG}\nADD app.tar.gz",
```
tmds: There should be a space and a dot at the end: `FROM ${DOTNET_RUNTIME_IMAGE_STREAM_TAG}\nADD app.tar.gz .`

aslicerh: You are correct. Good catch!

aslicerh:
Do we want to wait and do this against master after the merge?

aslicerh:
I guess that determines that question. This PR has been changed to be against master.

jerboaa:
We'll merge this once images are available.

aslicerh:
I like that plan.

PR #103, opened 2017-08-10
Update dotnet-example template for 2.0

jerboaa:
We are going to merge this once 2.0 images will be available.

### DIFF
--- a/templates/dotnet-example.json
+++ b/templates/dotnet-example.json
@@ -249,7 +249,7 @@
             "displayName": ".NET builder",
             "required": true,
             "description": "The image stream tag which is used to build the code.",
-            "value": "dotnet:1.0"
+            "value": "dotnet:2.0"
         },
         {
             "name": "NAMESPACE",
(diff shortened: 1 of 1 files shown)


tmds:
Can you remove setting `DOTNET_PUBLISH` from the template?

tmds:
Include dotnet-pgsql-persistent.json in this PR too?

jerboaa:
> Include dotnet-pgsql-persistent.json in this PR too?

I'm not there, yet so will be a separate PR, since I've not tested it yet.

jerboaa:
> Can you remove setting DOTNET_PUBLISH from the template?

Do you mean removing this [environment variable](https://github.com/redhat-developer/s2i-dotnetcore/blob/master/templates/dotnet-example.json#L111)? Wouldn't this *not* publish if used like so:

    $ oc new-app --template=dotnet-example --param=DOTNET_IMAGE_STREAM_TAG=dotnet:1.1 --param=SOURCE_REPOSITORY_REF=dotnetcore-1.1

tmds:
I see, yes, when using the template to instantiate a 1.x project it would require the variable still.

jerboaa:
Once there are no active 1.x images any more we can remove it.

PR #104, opened 2017-08-10
Review permission assignment

tmds:
- ensure the NSS_WRAPPER_PASSWD is writable
- assemble: don't persist NSS_WRAPPER_PASSWD
- runtime/Dockerfile: use fix-permissions script to update permissions
the chmod instruction in the Dockerfile would add the execute permission to every file, the fix-permissions script will only add it to directories

### DIFF
--- a/2.0/build/s2i/bin/assemble
+++ b/2.0/build/s2i/bin/assemble
@@ -96,3 +96,6 @@ chmod +x "$DOTNET_APP_PATH/$DOTNET_DEFAULT_CMD"
 fix-permissions ./
 # set permissions for any installed artifacts
 fix-permissions /opt/app-root
+
+# Don't persist the nss_wrapper passwd file
+rm -rf $NSS_WRAPPER_PASSWD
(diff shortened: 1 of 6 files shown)


2.0/build/s2i/bin/assemble
```diff
@@ -96,3 +96,6 @@ chmod +x "$DOTNET_APP_PATH/$DOTNET_DEFAULT_CMD"
 fix-permissions ./
 # set permissions for any installed artifacts
 fix-permissions /opt/app-root
+
+# Don't persist the nss_wrapper passwd file
+rm -rf $NSS_WRAPPER_PASSWD
```
jerboaa: Why not persist it? At runtime there will be a new user added, to this file. Not sure what this is solving. No strong feelings, though.

tmds: Removed it. I also moved the pid check from the container-entrypoint to the generate_container_user script. This makes it more clear when and how we are generating the NSS_WRAPPER_PASSWD file.

PR #105, opened 2017-08-10
Add DOTNET_PACK to pack the app as tar.gz and test split-build

tmds:


tmds:
For the spit build, this means something like:
build image:
```
DOTNET_PACK=true
```
copy artifacts:
```
sourcePath: /opt/app-root/app.tar.gz -> destinationDir: .
```
Dockerfile
```
FROM dotnet-runtime:2.0
ADD app.tar.gz .
```

2.0/build/test/run
```diff
@@ -501,6 +502,56 @@ test_user() {
   info "Test user PASSED."
 }
 
+test_split_build() {
```
aslicerh: This doesn't really work for QE. The two images get tested independently and do not have access to each other.

So at the least this test needs to be optional.

jerboaa: This should be OK, Andrew, as the dep chain is runtime-> builder. So at the point when build is being built, the runtime image should be accessible. It would not work if the runtime tests depend on the builder.

2.0/build/test/run
```diff
@@ -501,6 +502,56 @@ test_user() {
...
+  # remove container
+  docker rm -v $id
+  # remove image
+  cleanup ${app}
+
+  # create image based on app.tar.gz
+  image="${image}-runtime"
+  cat >Dockerfile <<EOF
+FROM ${RUNTIME_IMAGE_NAME}
+ADD app.tar.gz .
```
jerboaa: Isn't this missing a `dotnet [some.dll]`? Perhaps not, but there is a lot of magic going on. This probably works, because there is a `defauld_cmd.sh` that'll work? If so, please add a comment how this is supposed to work as it'll be difficult to diagnose later on when something changes.

2.0/build/README.md
```diff
@@ -136,6 +136,10 @@ a `.s2i/environment` file inside your source code repository.
 
     Publish assuming the runtime image contains the ASP.NET Core runtime store. Defaults to `false`.
 
+* **DOTNET_PACK**
```
jerboaa: Suggested alternative name `DOTNET_PUBLISH_PACK`

2.0/build/test/run
tmds: Comment added

2.0/build/README.md
tmds: I have a slight preference for the shorter form.

PR #112, opened 2017-08-24
Refactor build tests per feature

tmds:
Looking at 'build/test/run' it's hard to figure out what we are testing.
This PR refactors the tests into features.

### DIFF
--- a/2.0/build/test/run
+++ b/2.0/build/test/run
@@ -16,599 +16,312 @@
 
 IMAGE_NAME=${IMAGE_NAME:-dotnet/dotnet-20-rhel7}
 RUNTIME_IMAGE_NAME=${RUNTIME_IMAGE_NAME:-dotnet/dotnet-20-runtime-rhel7}
-
 OPENSHIFT_ONLY=${OPENSHIFT_ONLY:-false}
 
-if [ "$DEBUG" != "" ]; then
-  set -x
-fi
-
-declare -a CLI_APPS=(
-  hw_framework_config
-  helloworld
-  helloworld-fs
-  qotd
-  dotnet-configuration-debug
-  dotnet-configuration-release
-)
-declare -a WEB_APPS=(
-  asp-net-hello-world
-  asp-net-hello-world-envvar
-  asp-net-hello-world-envvar2
-)
-
 test_dir="$(readlink -zf $(dirname "${BASH_SOURCE[0]}"))"
-image_dir=$(readlink -zf ${test_dir}/..)
+source ${test_dir}/../../testcommon
+
 dotnet_version="2.0.0"
 sample_app_url="https://github.com/redhat-developer/s2i-dotnetcore-ex.git"
 
-test_port=8080
-
-info() {
-  echo -e "\n\e[1m[INFO] $@\e[0m\n"
+s2i_image_tag() {
+  local app="$1"
+  echo "${IMAGE_NAME}-${app}"
 }
 
-image_exists() {
-  docker inspect $1 &>/dev/null
+s2i_build_output_log() {
+  local app="$1"
+  local tag="$2"
+
+  docker_rmi ${tag}
+  s2i build --force-pull=false "file://${test_dir}/${app}" ${IMAGE_NAME} ${tag} 2>&1
 }
 
-container_exists() {
-  local cid_file="$1"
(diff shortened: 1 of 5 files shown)


tmds:
CI is failing
```
{"message":"open /var/lib/docker/image/devicemapper/layerdb/tmp/layer-000421683/diff: no space left on device"}
```

aslicerh:
When we 'respin' or create a container, we do it by copying the relevant sub-directories into the local RH container repos. This means everything in the 'build' folder gets copied into a repo and everything in the 'runtime' folder gets copied into a completely separate repo.

When we give the containers to QE, each repo has to be tested independently.

By having a 'common tests' script like that we would have to copy that script to each repo and change the test file to source it locally. That seems like a lot of work and a lot of room for error to achieve a little bit of code deduplication in a test script.

tmds:
I wasn't aware. So either we duplicate the file, or we make them symlinks and the cp is done with the L flag. Any preference?

2.0/build/test/testcommon
```diff
@@ -0,0 +1,180 @@
+# This code is the same for the build and runtime images.
+# The code is duplicated in both folders because the images are built
+# and tested after copying the sub-directories into a container repo.
+
+test_port=8080
+CURLE_COULNDT_CONNECT=7
```
jerboaa: Typo: `CURL_COULDNT_CONNECT`

2.0/build/test/run
```diff
@@ -16,599 +16,333 @@
...
-  done
-
-  for app in ${WEB_APPS[@]}; do
-    test_web_application "${app}"
-  done
-
-  test_new_web_app
-  test_user
-  test_split_build
+if [ ${OPENSHIFT_ONLY} != true ]; then
```
jerboaa: Could we add a test for the `true` case, which builds s2i-dotnetcore-ex for the relevant branch for 1.0, 1.1 and 2.0 respectively? The sample app should never break. We reference it in the imagestream. I'm fine with doing this in a separate patch/PR. Whatever you prefer.

2.0/runtime/test/run
```diff
@@ -12,154 +12,89 @@
 # Example usage: $ sudo ./test/run
 
 IMAGE_NAME=${IMAGE_NAME:-dotnet/dotnet-20-runtime-rhel7}
+OPENSHIFT_ONLY=${OPENSHIFT_ONLY:-false}
```
jerboaa: For runtime we can probably ignore `OPENSHIFT_ONLY`?

2.0/runtime/test/run
```diff
@@ -12,154 +12,89 @@
...
   assert_equal "${response}" "Hello World!"
 
-  popd
+  popd >/dev/null
 }
-test_aspnet
+
+info "Testing ${IMAGE_NAME}"
+
+if [ ${OPENSHIFT_ONLY} != true ]; then
```
jerboaa: I'd run those unconditionally and ignore OPENSHIFT_ONLY for runtime.

2.0/runtime/test/testcommon
```diff
@@ -0,0 +1,180 @@
+# This code is the same for the build and runtime images.
+# The code is duplicated in both folders because the images are built
+# and tested after copying the sub-directories into a container repo.
+
+test_port=8080
+CURLE_COULNDT_CONNECT=7
```
jerboaa: TYPO: `CURL_COULDNT_CONNECT`

2.0/build/test/testcommon
tmds: This is how the constant is named in libcurl (https://linux.die.net/man/3/libcurl-errors). The E is for error.

2.0/build/test/run
tmds: Test added

2.0/runtime/test/run
tmds: I removed this, so now all runtime tests run when OPENSHIFT_ONLY=true.

tmds:
@jerboaa thank you for reviewing. I've addressed the comments.

2.0/build/test/testcommon
jerboaa: I see.

Issue #113, opened 2017-08-25
Permission issues using Red Hat Developer Studio


mickaelistria:
As Red Hat Developer Studio enables easy continuous deploy with `rsync`, it would be great to include `rsync` in the container to allow continuous deployment from this IDE.

mickaelistria:
Here is the error I see in Red Hat Developer Studio trying to deploy to the dotnet-2.0 image
```
Error while publishing server dotnet-demo-2 (Service) at OpenShift 3 (console.starter-us-east-2.openshift.com).  Could not sync all pods to folder /home/mistria/devstudio-11.0.0.GA/workspace/.metadata/.plugins/org.jboss.ide.eclipse.as.core/dotnet-demo@dotnet-demo-2/deploy
Syncing com.openshift.restclient.capability.resources.IRSyncable$PodPeer@d744125 to com.openshift.restclient.capability.resources.IRSyncable$LocalPeer@76ba4fa4 failed: WARNING: cannot use rsync: rsync not available in container
ERROR: Error extracting file "./.git/objects/08/97bec7f077256f459646731bca3319b66e58ed": open /home/mistria/devstudio-11.0.0.GA/workspace/.metadata/.plugins/org.jboss.ide.eclipse.as.core/dotnet-demo@dotnet-demo-2/deploy/.git/objects/08/97bec7f077256f459646731bca3319b66e58ed: permission denied
ERROR: Error extracting tar stream
Ignoring the following flags because they only apply to rsync: --exclude
error: error extracting tar at destination directory: open /home/mistria/devstudio-11.0.0.GA/workspace/.metadata/.plugins/org.jboss.ide.eclipse.as.core/dotnet-demo@dotnet-demo-2/deploy/.git/objects/08/97bec7f077256f459646731bca3319b66e58ed: permission denied
```
@jeffmaury @bdshadow Do you think the issue is in the image or in RHDS?

mickaelistria:
I found the issue I got was caused by the project being at the root of the Git repository. I reorganized stuff, and the issue I see now is
```
Could not sync /home/mistria/devstudio-11.0.0.GA/workspace/.metadata/.plugins/org.jboss.ide.eclipse.as.core/dotnet-demo@demo-dotnet/deploy to all pods running the service demo-dotnet
[...]

PR #114, opened 2017-08-25
Run fix-permissions at the end of the 2.0 build dockerfile.

aslicerh:


aslicerh:
See the comment in the code diff for what this does.

I hit this issue rebuilding the 2.0 containers for the F# changes. The s2i assembly script would fail because it tried to run fix-permissions on /opt/app-root and would fail on a '.pki' directory that contained an nssdb. This looks like it didn't fail previously because that directory was created in the runtime image instead, so it received its permissions properly.

So yeah, call this a little fool-proofing in case future package installs create more directories like this.

tmds:
> I hit this issue rebuilding the 2.0 containers for the F# changes.

I'm not seeing an issue when I build master. How do you observe the issue?

2.0/build/Dockerfile.rhel7
```diff
@@ -33,6 +33,11 @@ RUN INSTALL_PKGS="rh-nodejs6-npm rh-dotnet20-dotnet-sdk-2.0" && \
     rpm -V $INSTALL_PKGS && \
     yum clean all -y
 
+# Since $HOME is set to /opt/app-root, the yum install may have created config
```
jerboaa: This refers to `yum install` without specifics. Perhaps mention: "`~/.pki/nssdb` for example". It seems to be used by nss (and some other programs): https://wiki.mozilla.org/NSS_Shared_DB_And_LINUX. That would explain in a bit more detail what artifacts to look for further down the road.

tmds:
Do you have a different output when running:
```
$ docker run dotnet/dotnet-20-rhel7 ls -lah /opt/app-root/.pki
total 0
drwxrwxrw-. 3 default root 19 Aug 28 08:53 .
drwxrwxrwx. 6 default root 51 Aug 28 08:56 ..
drwxrwxrw-. 2 default root  6 Aug 28 08:53 nssdb
```

2.0/build/Dockerfile.rhel7
```diff
@@ -33,6 +33,11 @@ RUN INSTALL_PKGS="rh-nodejs6-npm rh-dotnet20-dotnet-sdk-2.0" && \
...
+# directories there. These will be owned by root and can cause actions that work
+# on all of /opt/app-root to fail. So we need to fix the permissions on those too.
+RUN chown -R 1001:0 /opt/app-root && fix-permissions /opt/app-root
```
tmds: @aslicerh Can you move `mkdir /opt/app-root/src` from below to here, so that folder also has proper permissions.

aslicerh:
@tmds What did it for me was trying to build the 2.0 build image using the 2.0-4 runtime image as the base. The problem is consistent for me, but why don't you try to reproduce it? Change the FROM line in the 2.0 build dockerfile to have `dotnet/dotnet-20-runtime-rhel7:2.0-4`. Do a docker build using it.

That new image should have a .pki directory in /opt/app-root that is owned by root.

2.0/build/Dockerfile.rhel7
aslicerh: That sounds like a good idea.

2.0/build/Dockerfile.rhel7
aslicerh: Sure, I'll add a note about nssdb to the comment. But the reason I referred to this as 'future proofing' above is because we can't predict what kinds of things our dependencies may create, not just because of the nssdb.

Though I can understand wanting to add a general example.

tmds:
> What did it for me was trying to build the 2.0 build image using the 2.0-4 runtime image as the base. The problem is consistent for me, but why don't you try to reproduce it? Change the FROM line in the 2.0 build dockerfile to have dotnet/dotnet-20-runtime-rhel7:2.0-4. Do a docker build using it.

It looks like the published images don't have this file This explains why the issue shows up when you changed the base image.
```
$ docker run registry.access.redhat.com/dotnet/dotnet-20-runtime-rhel7:2.0 ls -lah /opt/app-root/.pki
ls: cannot access /opt/app-root/.pki: No such file or directory
$ docker run registry.access.redhat.com/dotnet/dotnet-20-rhel7:2.0 ls -lah /opt/app-root/.pki
ls: cannot access /opt/app-root/.pki: No such file or directory
```

aslicerh:
Alright, sorry for the wait, but here are updates related to the comments so far.

@tmds That would seem to be correct, yes. When I build them myself the images show up. I was trying to look into that some last week but I am not sure what causes the difference between my VM and the build server though.

aslicerh:
This should also address the permissions issues with the src directory seen in Issue #113.

PR #115, opened 2017-08-28
Change default s2i_args

asaf400:
Since s2i v1.0.4 the '--force-pull=false' argument has been deprecated (see release 1.0.4 @ https://github.com/openshift/source-to-image/releases/tag/v1.0.4)

and since s2i v1.1.6, the '--force-pull=false' argument has been removed completely and users should now use only '--pull-policy' argument ( see release 1.1.6 @ https://github.com/openshift/source-to-image/releases/tag/v1.1.6)

These 4 commits change the args in each image version to support the newer s2i releases from 1.0.4 and later..

### DIFF
--- a/1.0/test/run
+++ b/1.0/test/run
@@ -438,7 +438,7 @@ info "Testing ${IMAGE_NAME}"
 
 # Since we built the candidate image locally, we don't want S2I attempt to pull
 # it from Docker hub
-s2i_args="--force-pull=false"
+s2i_args="--pull-policy=never"
 
 # Verify the 'usage' script is working properly when running the base image with 's2i usage ...'
 test_s2i_usage "${sample_app_url}"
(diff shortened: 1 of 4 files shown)


tmds:
My subscribed rhel 7.4 has 1.0.9-1.el7 as latest, so this feature is available.
@asaf400 Can you remove the changes to the 2.0 branch, so I don't have to deal with merge conflicts on: https://github.com/redhat-developer/s2i-dotnetcore/pull/112? I will include these changes on that PR.
@jerboaa can you have a look and merge if ok?
Thank you!

jerboaa:
[cibuild]

jerboaa:
@tmds This looks OK to me. I'd also suggest to keep this patch as is and resolve conflicts in #112 once it's closer to a solution.

tmds:
> @tmds This looks OK to me. I'd also suggest to keep this patch as is and resolve conflicts in #112 once it's closer to a solution.
ok, I'll deal with the merge conflicts.

Issue #71, opened 2017-05-01
support debugging


tanaka-takayoshi:
The current s2i dotnetcore docker image is available for remote debugging.But it requires a little bit complicated steps. I summarized the steps for remote debugging from Visual Studio code.
https://github.com/tanaka-takayoshi/s2i-dotnetcore/wiki/Remote-Debugging-s2i-dotnetcore-from-Visual-Studio-Code

Do you have a plan for making it easier? 

- pre-installed vsdbg in docker image
- pre-installed unzip in docker image (vsdbg is not installed by default, but make it easy to install vsdbg)

tmds:
@tanaka-takayoshi I have tried this before. This worked when I did it on a local OpenShift instance. It did not work when running on OpenShift online (see https://github.com/tmds/repro-sehexception). Does this work for you when running on OpenShift online?
The debugger has a very specific license. If we consider to include this, we must be sure that is permitted by the license.

tanaka-takayoshi:
@tms I just test my steps on OpenShift online preview and found it works well. I'm sorry I have no idea why your steps don't work well right now.
As of the license, I understand the situation. I should clarify the license of vsdbg at first.

tanaka-takayoshi:
The license of vsdbgs is described [here](https://www.visualstudio.com/license-terms/mt644895/). It appears publish with docker image is not allowed.

tmds:
@tanaka-takayoshi can you update your procedure using the dotnet-runtime 2.0 image (registry.access.redhat.com/dotnet/dotnet-20-runtime-rhel7:2.0) and see what packages are missing?

To use the runtime image, you need to:
- do build an image yourself with pre-compiled binaries (https://access.redhat.com/documentation/en-us/net_core/2.0/html/getting_started_guide/gs_install_dotnet#run_applications_on_docker);
- or do a chained build (https://access.redhat.com/documentation/en-us/net_core/2.0/html/getting_started_guide/gs_dotnet_on_openshift#create_runtime_image).

tanaka-takayoshi:
@tmds I have just confirmed the procedure using the dotnet-runtime 2.0 image. I'll update procedure soon.

``unzip`` command is required to install vsdbg. Also, ``rsync`` helps to transfer files.

I need below workaround if unzip is not enabled.
```
$ curl -sSL https://aka.ms/getvsdbgsh | bash /dev/stdin -v latest -l ./vsdbg -r linux-x64
[...]

PR #120, opened 2017-09-07
Add unzip to runtime image for easy install debugger

tmds:
Implements https://github.com/redhat-developer/s2i-dotnetcore/issues/71

CC @tanaka-takayoshi


tmds:
Thanks @jerboaa . I've added back the OPENSHIFT_ONLY check to the runtime tests because it was trying to find unzip in the registry.access.redhat.com images causing CI to fail.

PR #122, opened 2017-09-11
Move rsync from runtime image to build image.

aslicerh:
After the merge of PR #119 there was expressed some more concern that the addition of the rsync package to the runtime had not been asked for yet, which would make the additional size unnecessary. So this moves it to the build image until such a time as more users express a need for it.


tmds:
Thanks Andrew!

PR #134, opened 2017-10-16
Add CentOS based images for 2.0

tmds:
Implements https://github.com/redhat-developer/s2i-dotnetcore/issues/36

### DIFF
--- a/1.0/test/run
+++ b/1.0/test/run
@@ -33,7 +33,7 @@ declare -a WEB_APPS=(asp-net-hello-world asp-net-hello-world-envvar)
 
 test_dir="$(readlink -zf $(dirname "${BASH_SOURCE[0]}"))"
 image_dir=$(readlink -zf ${test_dir}/..)
-dotnet_version="1.0.0-preview2-003159"
+dotnet_version="1.0.0-preview2-003200"
 sample_app_url="https://github.com/redhat-developer/s2i-dotnetcore-ex.git"
 
 test_port=8080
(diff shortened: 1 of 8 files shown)


tmds:
I've included the following in the Jenkins pr-checker config: 
```
# Build and test CentOS images
TEST_OPENSHIFT=false BUILD_CENTOS=true ./build.sh`
```

jerboaa:
This looks good to me. Nice work!

As a follow-up we might want to create a CentOS mirror of this too:
https://github.com/redhat-developer/dotnet-jenkins-slave

jerboaa:
Then we could also add a mirror of the [image stream](https://github.com/redhat-developer/s2i-dotnetcore/blob/master/dotnet_imagestreams.json) and the templates should work with centos versions too.

aslicerh:
Can we name the new Dockerfile to something like `Dockerfile.centos`? That would keep the naming convention consistent, and it would get in the way less for the container builds, where we currently soft link `Dockerfile.rhel7` to `Dockerfile`.

tmds:
> Can we name the new Dockerfile to something like Dockerfile.centos?

I named this `Dockerfile` since it seems to be the naming convention for CentOS in the other s2i repos (e.g. [nodejs](https://github.com/sclorg/s2i-nodejs-container/tree/master/8), [python](https://github.com/sclorg/s2i-python-container/tree/master/3.6)).

@jerboaa wdyt?

jerboaa:
> Can we name the new Dockerfile to something like Dockerfile.centos?

What @tmds  the convention is Dockerfile being the centos version upstream. Dockerfile.rhel7 being the RHEL-based version. We should stick with that convention as it will make something like this work out-of-the-box on OpenShift:

     $ oc new-build --strategy=docker --context-dir=2.0/runtime https://github.com/redhat-developer/s2i-dotnetcore

tmds:
Thank you for review @jerboaa @aslicerh!

PR #137, opened 2017-10-18
Use dotnet prefix for CentOS images.

jerboaa:
The assumption was that the prefix in the CentOS registry will be `centos`.
But it isn't. It's `dotnet` as for RHEL. The main distinction is the `centos7` suffix.

### DIFF
--- a/2.0/build/Dockerfile
+++ b/2.0/build/Dockerfile
@@ -1,4 +1,4 @@
-FROM centos/dotnet-20-runtime-centos7
+FROM dotnet/dotnet-20-runtime-centos7
 # This image provides a .NET Core 2.0 environment you can use to run your .NET
 # applications.
 
@@ -14,7 +14,7 @@ LABEL io.k8s.description="Platform for building and running .NET Core 2.0 applic
       io.s2i.scripts-url=image:///usr/libexec/s2i
 
 # Labels consumed by Red Hat build service
-LABEL name="centos/dotnet-20-centos7" \
+LABEL name="dotnet/dotnet-20-centos7" \
       com.redhat.component="rh-dotnet20-docker" \
       version="2.0" \
       release="1" \
--- a/2.0/build/test/run
+++ b/2.0/build/test/run
@@ -15,8 +15,8 @@
 # Example usage: $ sudo ./test/run
 
 if [ "$BUILD_CENTOS" = "true" ]; then
-  IMAGE_NAME=${IMAGE_NAME:-centos/dotnet-20-centos7}
-  RUNTIME_IMAGE_NAME=${RUNTIME_IMAGE_NAME:-centos/dotnet-20-runtime-centos7}
+  IMAGE_NAME=${IMAGE_NAME:-dotnet/dotnet-20-centos7}
+  RUNTIME_IMAGE_NAME=${RUNTIME_IMAGE_NAME:-dotnet/dotnet-20-runtime-centos7}
 else
   IMAGE_NAME=${IMAGE_NAME:-dotnet/dotnet-20-rhel7}
   RUNTIME_IMAGE_NAME=${RUNTIME_IMAGE_NAME:-dotnet/dotnet-20-runtime-rhel7}
--- a/2.0/runtime/Dockerfile
+++ b/2.0/runtime/Dockerfile
@@ -19,7 +19,7 @@ LABEL io.k8s.description="Platform for running .NET Core 2.0 applications" \
       io.openshift.expose-services="8080:http"
 
 # Labels consumed by Red Hat build service
-LABEL name="centos/dotnet-20-runtime-centos7" \
+LABEL name="dotnet/dotnet-20-runtime-centos7" \
       com.redhat.component="rh-dotnet20-runtime-docker" \
       version="2.0" \
       release="1" \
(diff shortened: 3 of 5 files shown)


jerboaa:
@omajid @tmds This together with https://github.com/CentOS/container-index/pull/215 should get the builder image working I'd hope. Thoughts?

tmds:
Why are we the only s2i that has their own yml file in centos index?

jerboaa:
> Why are we the only s2i that has their own yml file in centos index?

Are we? sclo has some too no?
https://github.com/CentOS/container-index/blob/master/index.d/sclo.yml#L193

tmds:
I don't know what sclo means,  the yml file seems to collect a bunch of s2is, why doesn't it contain ours?

jerboaa:
It appears it's all images maintained in https://github.com/sclorg/ Not sure whether or not there is a convention anymore. It makes sense for our images to be in one file. I suppose the jenkins slave will get added there too.

tmds:
sclorg is related to Red Hat.
We are the first s2i not maintained by sclorg:
```
$ grep -r '/s2i' | sort | uniq
index.d/dotnet.yml:    git-url: https://github.com/redhat-developer/s2i-dotnetcore
index.d/sclo.yml:    git-url: https://github.com/sclorg/s2i-nodejs-container
index.d/sclo.yml:    git-url: https://github.com/sclorg/s2i-perl-container
index.d/sclo.yml:    git-url: https://github.com/sclorg/s2i-php-container
index.d/sclo.yml:    git-url: https://github.com/sclorg/s2i-python-container
index.d/sclo.yml:    git-url: https://github.com/sclorg/s2i-ruby-container
```

tmds:
> We are the first s2i not maintained by sclorg:

By that I mean: we are the first s2i not using the sclorg prefix.

tmds:
Who is maintaining the images on DockerHub under centos.
e.g. from https://github.com/sclorg/s2i-python-container

> This image is available on DockerHub. To download it run:
$ docker pull centos/python-36-centos7

python-36-centos7 is not in the CentOS/container-index :confused:

tmds:
> centos/python-36-centos7

It's based on seeing this that I adopted that convention our repo.
CentOS/container-index actually doesn't care how we name this in our repo. It just picks up our Dockerfile and than tags it based on the yml file.

tmds:
> CentOS/container-index actually doesn't care how we name this in our repo. It just picks up our Dockerfile and than tags it based on its yml file.

Actually, it will fail probably on the FROM in the build image?
If we want to put these images on DockerHub under centos/ probably, this means we should put them under centos.yml in CentOS/container-index?

tmds:
Answering my own questions:

> If we want to put these images on DockerHub under centos/ probably, this means we should put them under centos.yml in CentOS/container-index?

So 'sclo.yml' images:
- are built for registry.centos.org
- are built again for hub.docker.com (specific jenkins job for sclo)

A user can use them as 'centos/<image>' or as 'registry.centos.org/sclo/<image>'.

> Actually, it will fail probably on the FROM in the build image?

If the images are on registry.centos.org only (dotnet.yml), the FROM needs to be registry.centos.org/dotnet/...
Even if we'd put them in centos.yml, FROM centos/... would still mean: get them on docker hub.

jerboaa:
> If the images are on registry.centos.org only (dotnet.yml), the FROM needs to be registry.centos.org/dotnet/...
Even if we'd put them in centos.yml, FROM centos/... would still mean: get them on docker hub.

Can you clarify what you see the issue is with this post merge of this PR? In my book we don't need the registry prefix and we perhaps shouldn't add it, since runtime and builder are a unit and they should be built together.

From the centos pipeline library build log we have this:

    Sending build context to Docker daemon 135.2 kB

    Step 1 : FROM centos/dotnet-20-runtime-centos7
    Trying to pull repository c2bf.rdu2.centos.org/centos/dotnet-20-runtime-centos7 ... 
    Trying to pull repository registry.centos.org/centos/dotnet-20-runtime-centos7 ... 
    Trying to pull repository docker.io/centos/dotnet-20-runtime-centos7 ... 
    unauthorized: authentication required

Thus, it'll try registry.centos.org/dotnet/dotnet-20-runtime-centos7 which will be there and the build passes. Thoughts?

jerboaa:
[cibuild]

tmds:
> since runtime and builder are a unit and they should be built together.

I also consider them a unit. I've gotten the feedback before we build them separate.

> FROM centos/... would still mean: get them on docker hub.

The official docker daemon doesn't allow you to change the default registry (which is docker hub). Seems the one on the CI server is doing that, I wasn't aware.
With that lookup, things can build.

jerboaa:
@tmds OK to merge (provided CI passes)?

jerboaa:
> I also consider them a unit. I've gotten the feedback before we build them separate.

Let's describe what I mean by unit. When you build runtime/builder locally you are building both, but in two separate steps. The unit argument is that the sources to build from for both should be the same. Same upstream sha. So whenever runtime is being built, builder needs to be built too and vice versa.

Downstream does the same.

tmds:
> Downstream does the same.

I imagine using a similar setup with the docker registries.

> @tmds OK to merge (provided CI passes)?

You didn't catch the 'centos' prefix mention in the README.
ok to merge.

This means we are the first s2i builder officially (and exclusively) on registry.centos.org. :tada:

PR #139, opened 2017-10-20
Fix README for CentOS images.

jerboaa:
PR https://github.com/redhat-developer/s2i-dotnetcore/pull/137 missed this
as @tmds pointed out. This should fix it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,14 +44,13 @@ see [1.0 usage documentation](1.0/README.md).
 Image name structure
 ------------------------
 
-##### Structure: 1/2-3-4
+##### Structure: dotnet/1-2-3
 
-1. Prefix: 'dotnet' for RHEL, 'centos' for CentOS
-2. Platform name: 'dotnetcore' for 1.x, 'dotnet' for 2.0+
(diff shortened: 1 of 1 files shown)


jerboaa:
[skipci]

README.md
```diff
@@ -46,12 +46,12 @@ Image name structure
 
 ##### Structure: 1/2-3-4
```
tmds: fyi, before I added centos prefix, it said: dotnet/1-2-3

suggestion: we can get rid of the numbers, and include the runtime: `dotnet/<prefix>-<version>-[runtime-]<os>`

tmds: I don't have a strong preference, so feel free to keep it as-is.

jerboaa:
This seems simpler, so I've updated it. Removed the `builder` in section "Base image". It's really there to stress the base stack, not whether it's a s2i builder or not.

jerboaa:
@tmds Good to go?

tmds:
:shipit:

jerboaa:
Thanks for the review!

Issue #140, opened 2017-10-23
HTTP_PROXY envvar not used by .NET Core


tmds:
.NET Core uses libcurl. libcurl picks up http_proxy (lower-case) but not HTTP_PROXY (upper-case). For other proxy envvars (e.g. HTTPS_PROXY), both the lower and upper case are used.

libcurl: https://github.com/curl/curl/blob/2de63ab179eb78630ee039ad94fb2a5423df522d/lib/url.c#L5203-L5214

jerboaa:
@tmds I don't get it. Curl sources seem to suggest it's using `HTTP_PROXY` (the upper case variant of `http_proxy`) iff `http_proxy` is not set.
https://github.com/curl/curl/blob/2de63ab179eb78630ee039ad94fb2a5423df522d/lib/url.c#L5217
https://github.com/curl/curl/blob/2de63ab179eb78630ee039ad94fb2a5423df522d/lib/url.c#L5185

Thoughts?

tmds:
@jerboaa it is trying uppercase for all things (e.g. all_proxy, https_proxy) except for http_proxy,
```
  /*
   * We don't try the uppercase version of HTTP_PROXY because of
   * security reasons:
   *
   * When curl is used in a webserver application
   * environment (cgi or php), this environment variable can
   * be controlled by the web server user by setting the
   * http header 'Proxy:' to some value.
   *
   * This can cause 'internal' http/ftp requests to be
   * arbitrarily redirected by any external attacker.
   */
  if(!prox && !strcasecompare("http_proxy", proxy_env)) {
    /* There was no lowercase variable, try the uppercase version: */
    Curl_strntoupper(proxy_env, proxy_env, sizeof(proxy_env));
    prox = curl_getenv(proxy_env);
  }
```

jerboaa:
Right, sorry. I somehow read the `!strcasecompare()` condition wrong. Makes sense now. Thanks.

PR #141, opened 2017-10-24
Use HTTP_PROXY in .NET Core

tmds:
Implements https://github.com/redhat-developer/s2i-dotnetcore/issues/140

I've implemented this in container-entrypoint so it applies also to the runtime image.
This makes the split build pick up the proxy, which in OpenShift is typically upper-case (https://docs.openshift.com/enterprise/3.0/admin_guide/http_proxies.html).

After review of 2.0 changes, I will apply the same changes to the 1.0 and 1.1 images.

### DIFF
--- a/1.0/root/usr/bin/container-entrypoint
+++ b/1.0/root/usr/bin/container-entrypoint
@@ -1,5 +1,14 @@
 #!/bin/bash
 
-set -eu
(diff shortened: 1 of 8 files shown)


tmds:
Thanks @jerboaa !

PR #142, opened 2017-10-28
Dockerfile for dotnet 1.0 on Centos 7

yudhiwidyatama:
With this we could run dotnet 1.0 on centos 7 OS, useful when RHEL subscription aren't easily available.

### DIFF
--- a/1.0/Dockerfile
+++ b/1.0/Dockerfile
@@ -1,0 +1,101 @@
+FROM centos:7
+# This image provides a .NET Core 1.0 environment you can use to run your .NET
+# applications.
+
+EXPOSE 8080
+
+ENV DOTNET_CORE_VERSION=1.0
+# Default to UTF-8 file.encoding
+ENV LANG=C.UTF-8 \
+    HOME=/opt/app-root \
+    PATH=/opt/app-root/src/.local/bin:/opt/app-root/src/bin:/opt/app-root/bin:/opt/app-root/node_modules/.bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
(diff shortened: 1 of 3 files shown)


tmds:
I'm curious. Have you considered migrating to 2.0? There is already a centos image, and it moves you away from the deprecated project.json format?

yudhiwidyatama:
Currently we contracted an external team to create an application, which unfortunately using dotnet 1.0, and nearing the end period of the contract. The Team lead stated that due to resource constraints and volume of changes needed they were unable to port it during this contract term.

jerboaa:
Could you please update the tests to be able to run with the resulting image please?

https://github.com/redhat-developer/s2i-dotnetcore/blob/master/1.0/test/run#L17
and
https://github.com/redhat-developer/s2i-dotnetcore/blob/master/build.sh#L83

yudhiwidyatama:
ok, I updated the 1.0/test/run and build.sh in revision 1a3b7e3

jerboaa:
[cibuild]

aslicerh:
This looks good to me as well, so I'm going to go ahead and merge it in.

jerboaa:
@yudhiwidyatama Thanks for the contribution!

If you'd like the image to be auto-built in the centos registry, here would be the steps in order to get this done:

1. Add a `cccp.yml` file similar to [this](https://github.com/redhat-developer/s2i-dotnetcore/blob/master/2.0/build/cccp.yml) for `1.0` folder. Suggested job id `dotnetcore-10-centos7`
2. Add an entry to the centos container catalog, file [dotnet.yml](https://github.com/CentOS/container-index/blob/master/index.d/dotnet.yml) for the `dotnet/dotnetcore-10-centos7` image.

After that's in, it should get built and pushed to the centos registry automatically.

Issue #146, opened 2017-11-09
Support including --no-restore in assemble dotnet publish command


dougtoppin:
The --no-restore argument in assemble dotnet publish can be helpful for more granular control of the publishing function. An example of this is a build on a disconnected network where the build host cannot reach a nuget server and all necessary nuget packages are either already on the build host or contained in the build image.

tmds:
Closing per discussion in https://github.com/redhat-developer/s2i-dotnetcore/pull/147: we haven't been able to reproduce the need for this flag or write a test that fails when not adding it.

Issue #145, opened 2017-11-09
Support setting dotnet publish verbosity level


dougtoppin:
For determining what is happening during the dotnet publish operation in assemble it would be helpful to be able to pass in a environment variable for setting verbosity level (the publish -v argument).

tmds:
Fixed by https://github.com/redhat-developer/s2i-dotnetcore/pull/156

PR #147, opened 2017-11-09
Enhancements for #146 and #145 to support publish --no-restore and verbosity

dougtoppin:
I found these changes helpful for controlling assemble dotnet publish to both add optional verbosity control and to support --no-restore for environments where an external repo such as nuget.org was not available.
These only apply to the dotnet 2.0 assemble script as they do not appear to be available in 1.0/1.1.

### DIFF
--- a/2.0/build/README.md
+++ b/2.0/build/README.md
@@ -147,14 +147,14 @@ a `.s2i/environment` file inside your source code repository.
     When set to `true`, the application restart automatically when the source code changes. `dotnet run`
     is used to start the application.
 
-* **DOTNET_NO_RESTORE**
+* **DOTNET_NO_RESTORE** (dotnet 2.0 only)
 
   Can be set to `--no-restore`, will be passed to the `dotnet publish` command in `assemble`. Intended to be used for
   situations where not attempting to connect to an external `nuget.org` site is useful. An example of this
   would be a disconnected build configuration where all desired nuget packages are already accessible to the
   build either on the host or in the image.
 
-* **DOTNET_VERBOSITY**
+* **DOTNET_VERBOSITY** (dotnet 2.0 only)
 
   Will be passed to `assemble` script `dotnet publish` command allowing additional build related descriptive output.
   Example values are `-v  [ q | n | d | diag ]` (for quiet, minimal, normal, detailed and diagnostic levels).
--- a/2.0/build/README.md
+++ b/2.0/build/README.md
@@ -147,6 +147,18 @@ a `.s2i/environment` file inside your source code repository.
     When set to `true`, the application restart automatically when the source code changes. `dotnet run`
     is used to start the application.
 
+* **DOTNET_NO_RESTORE**
+
+  Can be set to `--no-restore`, will be passed to the `dotnet publish` command in `assemble`. Intended to be used for
+  situations where not attempting to connect to an external `nuget.org` site is useful. An example of this
+  would be a disconnected build configuration where all desired nuget packages are already accessible to the
+  build either on the host or in the image.
+
+* **DOTNET_VERBOSITY**
+
+  Will be passed to `assemble` script `dotnet publish` command allowing additional build related descriptive output.
+  Example values are `-v  [ q | n | d | diag ]` (for quiet, minimal, normal, detailed and diagnostic levels).
+
 NPM
 ---
 
--- a/2.0/build/s2i/bin/assemble
+++ b/2.0/build/s2i/bin/assemble
@@ -91,8 +91,11 @@ done
 echo "---> Restoring application dependencies..."
 dotnet restore "$DOTNET_STARTUP_PROJECT" $RESTORE_OPTIONS
 echo "---> Publishing application..."
+
+echo "dotnet publish arguments: DOTNET_STARTUP_PROJECT=$DOTNET_STARTUP_PROJECT DOTNET_FRAMEWORK=$DOTNET_FRAMEWORK DOTNET_CONFIGURATION=$DOTNET_CONFIGURATION DOTNET_ASPNET_STORE=$DOTNET_ASPNET_STORE DOTNET_APP_PATH=$DOTNET_APP_PATH DOTNET_NO_RESTORE=$DOTNET_NO_RESTORE DOTNET_VERBOSITY=$DOTNET_VERBOSITY"
+
 dotnet publish "$DOTNET_STARTUP_PROJECT" -f "$DOTNET_FRAMEWORK" -c "$DOTNET_CONFIGURATION" \
-       --self-contained false /p:PublishWithAspNetCoreTargetManifest=$DOTNET_ASPNET_STORE -o "$DOTNET_APP_PATH"
+       --self-contained false /p:PublishWithAspNetCoreTargetManifest=$DOTNET_ASPNET_STORE -o "$DOTNET_APP_PATH" $DOTNET_NO_RESTORE $DOTNET_VERBOSITY
 
 # check if the assembly used by the script exists
 if [ ! -f "$DOTNET_APP_PATH/${APP_DLL_NAME}" ]; then
@@ -119,3 +122,4 @@ rm -rf ~/{.local,.nuget}
 
 # fix permissions
 fix-permissions /opt/app-root
+


jerboaa:
Thanks for the contribution!

Rather than creating two new environment variables which both essentially boil down to being passed to `dotnet publish`, I'd suggest to create one: `DOTNET_PUBLISH_OPTS`. Then setting this in `.s2i/environment` should be sufficient for your use case, no?

     # Do not restore, set verbosity to diag
     DOTNET_PUBLISH_OPTS="--no-restore -v diag"

Either way this will get implemented, please @dougtoppin add tests verifying this does the right thing.

I wonder, though, since we are doing an explicit `dotnet restore` right before the publish I wonder if `--no-restore` should be considered by default for `dotnet publish`. Is there a case where we'd need `dotnet publish` to do a restore in the s2i context?

@aslicerh @tmds Thoughts?

tmds:
> optional verbosity control

I think that's a nice feature.
I prefer a separate environment variable because I like to keep closed the options a user can set. If we open this up, a user can set an option, and if we decide to control the same option later, that can cause their build to fail.

The DOTNET_VERBOSE flag should be set on all commands, not just publish.
Also, if we are adding this, when set, I think we should print out all environment variables at the start of the assemble script (i.e. assemble becomes verbose too).

> support --no-restore for environments where an external repo such as nuget.org was not available

I don't understand. Why doesn't it fail on `dotnet restore`? Are you setting DOTNET_RESTORE_SOURCES?
I was under the assumption that `dotnet publish` would see the project was already restored and no longer do that.
If that is not the case, why doesn't this test fail (S2iDotNetCoreDummy isn't on nuget.org) : https://github.com/redhat-developer/s2i-dotnetcore/blob/3da9f567514f1203588ea47e112f57265e834027/2.0/build/test/asp-net-hello-world-envvar2/src/app/app.csproj#L24

tmds:
> since we are doing an explicit dotnet restore right before the publish I wonder if --no-restore should be considered by default for dotnet publish

That makes sense to me. Still I'd like to understand better why this is failing for @dougtoppin and not for our test.

jerboaa:
> >  optional verbosity control

> I think that's a nice feature.
> I prefer a separate environment variable because I like to keep closed the options a user can set. If we open this up, a user can set an option, and if we decide to control the same option later, that can cause their build to fail.

Sure, but then it needs to be implemented differently. The way @dougtoppin did it `DOTNET_NO_RESTORE="-v diag"` would work. So it can be set to any dotnet publish option or argument. Same for `DOTNET_VERBOSITY`

tmds:
> Sure, but then it needs to be implemented differently. The way @dougtoppin did it DOTNET_NO_RESTORE="-v diag" would work. So it can be set to any dotnet publish option or argument. Same for DOTNET_VERBOSITY

I think we are saying the same thing.
DOTNET_NO_RESTORE, I'm not sure we need this.
DOTNET_VERBOSITY, can be set to one of the msbuild verbosity values (q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic], ref: https://msdn.microsoft.com/en-us/library/ms164311.aspx).
When set
* the assemble script prints out all environment variables (let's just call `env | sort`).
* is passed as --verbose argument to every dotnet command

jerboaa:
I think so too. Just stressing that passing some variable verbatim calls for it being abused:

    dotnet publish $SOME_VAR

vs

    dotnet publish -v "$DOTNET_VERBOSE"

or some such.

dougtoppin:
Just catching up on the thread/replies. I will include more info (so this might be a little long).

The environment that I am is is OCP and I am using the registry.access.redhat.com/dotnet/dotnet-20-rhel7 base image which I created a custom image from. The custom image includes all of the nuget packages that I need because the build nodes do not have access to the Internet (and cannot reach nuget.org).

I am setting the DOTNET_RESTORE_SOURCES env var with it set to the packages directory (file:///...) in the custom image.

However, what I was seeing is that even with that set publish would still attempt to reach nuget.org at the end causing the build to fail in my disconnected environment.

After some investigation I stumbled on this `https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#nuget-defaults-file` and noted the statement When nuget.exe is run for the first time, NuGet.Config is generated and nuget.org is added as the default package source in the config file."

If I ran bash in a container by hand and did a dotnet publish in it I saw that publish would create NuGet.Config and include a reference to sources being at nuget.org. If NuGet.Config already existed it would append the nuget.org entry to the file. It would also create a tracking file in the NuGet.Config directory to indicate that it had been run.
This is what was causing it to try to reach nuget.org even with DOTNET_RESTORE_SOURCES set.

After seeing that behavior I tried the --no-restore argument to publish and that resulted in it running all the way through without trying to reach nuget.org at the end which is what I needed to have happen in my situation so I modified assemble in my custom image to do that and also include a way to optionally set verbosity to help awareness of what publish was doing.

I debated having a generic DOTNET_OPTIONAL_ARGS type thing which would allow anything to be set but figured it might be better to be more explicit on the use I needed.

Sorry if I did not do this in the best way. It was something that I needed in the near term and figured that passing along what I did as a PR might be helpful for a reference even if it was not the best implementation approach. I have not used s2i standalone (outside the container I mean) so I was also not sure if there were side effects in other environments with what I did.

If there is a better way to do this please go ahead and do that rather than using my PR or else I could do a new PR with a more appropriate implementation.

aslicerh:
@dougtoppin 
First, let me echo what has been said and thank you for your contribution! And no need to apologize, we're just trying to determine the best way to generalize your contribution here.

@all 
I ran the test suite with the dotnet publish line always using the `--no-restore` flag and everything seemed fine. Given that we manually restore, I don't see a problem with making that a permanent change.

tmds:
> After seeing that behavior I tried the --no-restore argument to publish and that resulted in it running all the way through

I'm not against adding `--no-restore` to the publish command. I wish I could reproduce this issue myself to better understand what we are fixing (or working around).

Some things I've tried (on rhel7):
* clearing nuget cache (~/.nuget/packages, ~/.local/share/NuGet); creating a new web app; doing restore; disconnecting from network; publishing (-> succeeds)
* clearing nuget cache (~/.nuget/packages, ~/.local/share/NuGet); create a new web app; run restore with non-existing --source (-> fails as expected); run dotnet restore (default to nuget, succeeds); run restore again with the non-existing --source (this no longer fails, which means that the tool knows the project is successfully restored already and doesn't need to contact the source)

If you can share a log of the failing dotnet publish with verbosity 'diag', that may help us find out why publish is trying to contact nuget.org.
If your project was already restored, the log should have something like:
```
         Running restore with 8 concurrent jobs.
         Reading project file /tmp/webapp/webapp.csproj.
         The restore inputs for 'webapp' have not changed. No further actions are required to complete the restore.
```
If it is trying to restore packages, the log looks like:
```
         Running restore with 8 concurrent jobs.
         Reading project file /tmp/webapp/webapp.csproj.
         Restoring packages for /tmp/webapp/webapp.csproj...
```

> Sorry if I did not do this in the best way. It was something that I needed in the near term and figured that passing along what I did as a PR might be helpful for a reference even if it was not the best implementation approach. I have not used s2i standalone (outside the container I mean) so I was also not sure if there were side effects in other environments with what I did.

That's fine. We are discussing to understand the root cause and to find the proper fix.

dougtoppin:
With the default verbosity level this is the tail of what was being logged when assemble was doing the dotnet publish before I changed to args to fix the problem.

...
Restore completed in 8.83 sec for /opt/app-root/src/demo-web-api/demo-web-api.csproj.
Generating MSBuild file /opt/app-root/src/demo-web-api/obj/demo-web-api.csproj.nuget.g.props.
Restore completed in 12.19 sec for /opt/app-root/src/demo-web-api/demo-web-api.csproj.
---> Publishing application...
/opt/rh/rh-dotnet20/root/usr/lib64/dotnet/sdk/2.0.0/NuGet.targets(102,5): error : Unable to load the service index for source https://api.nuget.org/v3/index.json. [/opt/app-root/src/demo-web-api/demo-web-api.csproj]
/opt/rh/rh-dotnet20/root/usr/lib64/dotnet/sdk/2.0.0/NuGet.targets(102,5): error :   An error occurred while sending the request. [/opt/app-root/src/demo-web-api/demo-web-api.csproj]
/opt/rh/rh-dotnet20/root/usr/lib64/dotnet/sdk/2.0.0/NuGet.targets(102,5): error :   Couldn't connect to server [/opt/app-root/src/demo-web-api/demo-web-api.csproj]
error: build error: non-zero (13) exit code from dougtoppin_dotnet-20-rhel7-custom@xxx

tmds:
> With the default verbosity level

@dougtoppin can you create a log with verbosity diag?

dougtoppin:
@tmds Yep, will do this weekend. I had the default verb level output handy so posted that yesterday.

dougtoppin:
It is possible that what might be contributing to the issue that I am having with the disconnected build is that I am not copying a completely populated /opt/app-root/.nuget/packages/ directory into the image. I see some overhead files in there that I was not including in the image.  I had been only copying in the package files themselves.
More to come when I understand this better.

tmds:
@dougtoppin can you give us an update?

dougtoppin:
@tmds I don't have access to the original disconnected environment with the issue so I am working on duplicating the problem by preventing my own test environment build instance from reaching nuget.org. I am hoping be able to reproduce the issue within the next week. I will also be able to more easily get more verbose output from it and will send it along when I have it.

larsmeldgaard:
Hi guys.
I'm trying to migrate a 1.1 project to 2.0 and when building in the s2i image i get this:
 ---> Publishing application...
/opt/rh/rh-dotnet20/root/usr/lib64/dotnet/sdk/2.0.0/NuGet.targets(102,5): error : Unable to load the service index for source https://api.nuget.org/v3/index.json.

Since you mentioned this particular error in your discussion I just wanted to know if this was a known error, or if theres a workaround or configuration setting I need to make in order to get the publish to work. The nuget itself works fine by adding the sources in environment variables.

Hope you can help a bit...

larsmeldgaard:
@dougtoppin I'm using an on-prem solution of Openshift that does not have access to internet. So it is very likely that this is causing errors. I have a local proxy-repo that i'm referencing during build and this works fine. However this connection to nuget is during publish phase (not the build phase). The call shouldn't even be there.
I have tried with a vanilla dotnet core project and this project does not have this problem. So maybe it is caused by some of the nuget packages I have or by something done in the migration part. I'm only speculating here, 'cause I got nothing... :-)

dougtoppin:
@larsmeldgaard What you are describing sounds like the same issue that I had. To resolve it, for the near term anyway, I had to modify assemble to include an optional argument (env var) where I could pass '--no-restore' to the dotnet publish command. This fixed the issue and I am still trying to figure out whether or not something in my customized dotnet 2.0 image is causing it to try to reach nuget.org even when a local nuget repo is provided.
I will try to spend some more time on reproducing the issue outside the environment that I was working in.

aslicerh:
@larsmeldgaard 
Since you're seeing the same issue, can you run your publish with 'diag' level verbosity and make the logs available to us?

aslicerh:
@larsmeldgaard 
There's no reason to get the latest from the repo yet. We haven't put in any of these changes at this moment because we were wanting to see why this is occurring first and haven't been able to recreate it yet.

You said that you are doing this through openshift and have an s2i image available. Do you have access to command line on one of the hosts, and do you have the privileges to run docker commands?

larsmeldgaard:
I don't have access for this, but maybe we should setup a session with my admins - I'm sure we can work together to try these things out.

tmds:
@larsmeldgaard If you can access the local-proxy that is used in OpenShift from your development machine, you should be able to reproduce it there.

If your machine can access api.nuget.org, then add it to your hosts file to disable that access.
Now you can run the instructions just as in the container.
You shouldn't forget to clean out your local packages cache.
Set the verbosity to diag and check the lines mentioned here: https://github.com/redhat-developer/s2i-dotnetcore/pull/147#issuecomment-343517442.

> I have tried with a vanilla dotnet core project and this project does not have this problem.

This is interesting. Try trimming out a failing project until it works. First remove the code. Then start removing libraries (first your company ones). Perhaps this leads you to a csproj file with only public packages that fails. Then we can reproduce the issue too.

tmds:
@dougtoppin @larsmeldgaard any progress in generating a log/reproducible case?

dougtoppin:
@tmds Since I do not have access to the original OpenShift disconnected environment with the issue I've been simulating it today by using s2i to do a build of a simple dotnet 2 app from the shell in an EC2 instance and using deny rules in the VPC NACL to block all access to nuget.org and api.nuget.org.
I have a local nuget server running in a container (sunside/simple-nuget-server) and have all of the packages that the app needs hosted in it.
I would expect that this configuration would duplicate the issue I had in the disconnected network, try to reach nuget.org and then fail but the build is completing successfully without including the do not restore argument to publish that I added to the custom image and the PR I sent in.

Because of this, assuming that I am simulating the configuration correctly, I must have had something in my custom image that was causing it to still try to reach the outside nuget server.
I have to admit I am not a pro in this area so I could be wrong but without being able to repro the issue in a controlled fashion I have to assume that it was me.
Sorry about that.

I do think that adding the ability to set the publish verbosity level in assemble can be helpful though.

Unless I stumble on the actual cause you can ignore my do not restore mods in the PR.

tmds:
@dougtoppin thanks for trying to reproduce.
I'm replacing this PR by https://github.com/redhat-developer/s2i-dotnetcore/pull/156 which only implements the verbosity feature.
If someone hits the publish-restore issue again, it will be easy for them to turn up the verbosity to get a detailed log.

PR #151, opened 2017-11-27
Sync with downstream container repos and bump release version in centos dockerfiles

aslicerh:
This should trigger a container rebuild in the centos registry, which should pull in the new dotnet release.


aslicerh:
@omajid 
The centos dotnet container pipeline definition file lists both a runtime and a build image, and gives the runtime as being a dependency of the build image.

The definition file can be found here: https://github.com/CentOS/container-index/blob/master/index.d/dotnet.yml

I'm not sure how smart the auto-build system is, so we may see two build containers get created. I'm not sure if that is going to be an issue or not yet.

aslicerh:
[cibuild]

aslicerh:
[cibuild]

aslicerh:
Ok, with the centos RPMs propagated now, it looks like everything is working again.

PR #156, opened 2018-01-08
Add DOTNET_VERBOSITY to specify verbosity of dotnet build commands

tmds:
Implements https://github.com/redhat-developer/s2i-dotnetcore/issues/145
Replaces https://github.com/redhat-developer/s2i-dotnetcore/pull/147

CC @dougtoppin

### DIFF
--- a/2.0/build/s2i/bin/assemble
+++ b/2.0/build/s2i/bin/assemble
@@ -2,6 +2,13 @@
 
 set -e
 
+DOTNET_OPTIONS=""
+if [ -n "${DOTNET_VERBOSITY}" ]; then
+  echo "--> Environment:"
+  env | sort
+  DOTNET_OPTIONS="$DOTNET_OPTIONS -v ${DOTNET_VERBOSITY}"
+fi
+
 # npm
 if [ -n "${DOTNET_NPM_TOOLS}" ]; then
   echo "---> Installing npm tools ..."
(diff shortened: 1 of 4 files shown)


2.0/build/s2i/bin/assemble
```diff
@@ -2,6 +2,13 @@
 
 set -e
 
+DOTNET_OPTIONS=""
+if [ -n "${DOTNET_VERBOSITY}" ]; then
+  echo "--> Environment:"
+  env | sort
+  DOTNET_OPTIONS="$DOTNET_OPTIONS -v ${DOTNET_VERBOSITY}"
```
jerboaa: If I'm reading this correctly, then DOTNET_OPTIONS will always be the empty string at this point. So perhaps we should rewrite this as:

     DOTNET_OPTIONS="-v ${DOTNET_VERBOSITY}"

Thoughts?

tmds:
> One question. Is there a chance of DOTNET_OPTIONS not being the empty string? If DOTNET_OPTIONS=foo prior line 5 in assemble, then it'll be reset to empty after it, no?

A user can set this. If they do, we want to discard their options. If we wouldn't discard them, they may set some options that seem to work fine. If we then later make some changes to assemble that also sets those options, that will break their build.

2.0/build/s2i/bin/assemble
tmds: I have a slight preference to always append options. This avoids accidental discarding options that were set previously.

jerboaa: Sure, but setting DOTNET_OPTIONS="" just prior the if branch will reset any user-setting. So perhaps you mean to do `DOTNET_OPTIONS=${DOTNET_OPTIONS:-}`?

jerboaa:
I'm a bit confused. In this [comment](https://github.com/redhat-developer/s2i-dotnetcore/pull/156#issuecomment-355966089) you say you'd like to discard user-set options. In that [comment](https://github.com/redhat-developer/s2i-dotnetcore/pull/156#discussion_r160143387) you seem to contradict it: "[...] Avoids accidental discarding of options". Which should it be?

tmds:
> I'm a bit confused. In this comment you say you'd like to discard user-set options. In that comment you seem to contradict it: "[...] Avoids accidental discarding of options". Which should it be?

For the end-user we don't want them to be able to set DOTNET_OPTIONS, since we cant grantee backwards compatibility, so we set it explicitly to "".

For our-selves, if we later use DOTNET_OPTIONS to add some additional flags, we need to make sure we don't get into this situation:
```
DOTNET_OPTIONS=""
+ if [ "$DOTNET_OTHERFLAG" != "true" ]; then
+   DOTNET_OPTIONS="--other-flag"
+ fi
if [ -n "${DOTNET_VERBOSITY}" ]; then
  DOTNET_OPTIONS="-v ${DOTNET_VERBOSITY}"
fi
```
Here DOTNET_VEBROSITY will cancel out DOTNET_OTHERFLAG.
For this reason I have a slight preference to append options (i.e. `DOTNET_OPTIONS="$DOTNET_OPTIONS ..."`). I can remove it if you prefer to that

2.0/build/s2i/bin/assemble
aslicerh: Speaking of appending for consistency, how about echoing the environment stuff if DOTNET_OPTIONS is not empty, instead of if DOTNET_VERBOSITY is specified.

That way if we add more options later, any of them being set will trigger the print.

tmds: > That way if we add more options later, any of them being set will trigger the print.

The rationale is to echo the environment when the build is made more verbose.

aslicerh: But it also prints if you set the verbosity to be less verbose.

tmds:
@jerboaa @aslicerh based on your feedback, I've replaced DOTNET_OPTIONS by VERBOSITY_OPTION.

tmds:
[cibuild]

aslicerh:
[cibuild]

tmds:
Thanks @jerboaa @aslicerh !

PR #161, opened 2018-02-27
Use 'Invariant' culture

tmds:
.NET Core determines the culture based on the LANG environment variable.
By setting to 'C.UTF-8', we were using the 'en-US-POSIX' culture.
By not setting LANG, .NET uses the 'Invariant' culture.

- Not setting LANG is consistent with Microsoft .NET Core images.
- Performance for the Invariant culture is better than the 'en-US-POSIX' culture.


tmds:
Technically this is a breaking change. Since the 'en-US-POSIX' culture is not common on workstation/images, in practice, this will be more a _not-breaking_ change. :smile:

aslicerh:
Yeah, I was concerned about the possibility of this breaking a customer configuration at first, but that situation isn't likely. If they needed support for something other than english, they probably already set their LANG in their image. If they use english, most people will use the invariant default. Very few people probably go out of their way to use english unicode.

So yeah, this works for me.

tmds:
Build failed with timeout:
```
++ '[' 7 -eq 7 -a 31 -le 30 ']'
++ '[' 7 -ne 0 ']'
++ error 'curl: (7) Failed connect to 172.17.0.2:8080; Connection refused'
++ echo -e '[ERROR] curl: (7) Failed connect to 172.17.0.2:8080; Connection refused'
[ERROR] curl: (7) Failed connect to 172.17.0.2:8080; Connection refused
```

tmds:
[cibuild]

tmds:
Thanks for reviewing @omajid @aslicerh

PR #167, opened 2018-03-28
Prettify openshift.io/display-names

tmds:
The display name shows up in lists with other s2i images which don't
include the words "Image"/"Builder".


tmds:
For example, this is catalog on openshift.com:

screenshot from 2018-03-28 15-59-33

aslicerh:
Looks good to me!

PR #173, opened 2018-04-18
dotnet-example: Fix DOTNET_STARTUP_PROJECT being replaced by DOTNET_SDK_VERSION.

tmds:
Regression in https://github.com/redhat-developer/s2i-dotnetcore/pull/154


tmds:
Thanks @aslicerh

PR #175, opened 2018-04-25
2.1: Remove DOTNET_ASPNET_STORE envvar

tmds:
.NET Core 2.1 does not have an ASP.NET Core store.
This mechanism was replaced by an ASP.NET Core shared framework.


tmds:
Thanks @aslicerh!

PR #177, opened 2018-04-27
Make DOTNET_TOOLS use DOTNET_RESTORE_SOURCES like DOTNET_STARTUP_PROJECT.

tmds:
`dotnet restore`'s --source option replaces global sources.
`dotnet tool install`'s --source-feed option is adding to global sources instead of replacing.
We can have consistent behavior by generating a NuGet config file that ignores global sources and passing that to `dotnet tool install`.


tmds:
> Just out of curiosity, did you actually see something not work correctly (like an older source being used instead of a newer one) or is this just for consistency?

If you are deploying this and you specify a local nuget server using DOTNET_RESTORE_SOURCES you don't want it to go and try to reach out to nuget.org.

aslicerh:
> If you are deploying this and you specify a local nuget server using DOTNET_RESTORE_SOURCES you don't want it to go and try to reach out to nuget.org.

I see. I was just curious what the error state looked like. Good to know, and good catch.

PR #179, opened 2018-05-22
Add explicit rm for yum cache to reduce image size.

aslicerh:
Turns out the `yum clear all -y` line doesn't clear out all the cache files. I believe this is because we disable the repos and then enable them manually. Explicitly removing the cache saves ~200-350 MB in the image, depending on the version.

Note: the centos images don't have this problem, assumidly because they don't disable repos like the rhel images do, but I went ahead and added the lines for consistency. If anyone disagrees with that, just let me know and I'll remove the changes to the centos images.

### DIFF
--- a/1.0/Dockerfile
+++ b/1.0/Dockerfile
@@ -34,6 +34,8 @@ COPY ./s2i/bin/ /usr/libexec/s2i
 # run and build the applications.
 COPY ./contrib/ /opt/app-root
 
+# The 'rm -rf /var/cache/yum' line is there to clean out orphaned yum cache files,
+# which can be quite large in size.
 RUN yum install -y centos-release-dotnet centos-release-scl-rh && \
     INSTALL_PKGS="rh-dotnetcore10 nss_wrapper tar rh-nodejs4-npm" && \
     yum install -y --setopt=tsflags=nodocs  \
--- a/1.0/Dockerfile
+++ b/1.0/Dockerfile
@@ -40,6 +40,7 @@ RUN yum install -y centos-release-dotnet centos-release-scl-rh && \
       $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all -y && \
+    rm -rf /var/cache/yum/* && \
     mkdir -p /opt/app-root/src /opt/app-root/publish && \
     useradd -u 1001 -r -g 0 -d /opt/app-root/src -s /sbin/nologin \
       -c "Default Application User" default && \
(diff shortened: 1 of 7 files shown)


1.0/Dockerfile
```diff
@@ -40,6 +40,7 @@ RUN yum install -y centos-release-dotnet centos-release-scl-rh && \
       $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all -y && \
+    rm -rf /var/cache/yum/* && \
```
tmds: Maybe add a comment why we're explicitly calling this.

aslicerh:
> > Explicitly removing the cache saves ~200-350 MB in the image

> Is that the uncompressed image after doing a local build? Or is it the squashed/compressed size we see in OpenShift?

That would be uncompressed.

1.0/Dockerfile
aslicerh: Will do.

1.0/Dockerfile
```diff
@@ -34,12 +34,15 @@ COPY ./s2i/bin/ /usr/libexec/s2i
 # run and build the applications.
 COPY ./contrib/ /opt/app-root
 
+# The 'rm -rf /var/cache/yum' line is there to clean out orphaned yum cache files,
+# which can be quite large in size.
```
tmds: nit: I think you can put that comment line just above the 'rm -rf' line and docker will happily ignore it. Then you don't have to refer to the line explicitly.
Give it a try and feel free to chose the style you like best.

aslicerh: That was how I had it at first, and it worked fine if the rm was the last line in the RUN directive, but if there were other lines after it then it didn't like it. I believe that this is because it executes them as if they are a single line, which comments out the rest of the instructions.

aslicerh: Nevermind, read your comment wrong. I tried inline, not a separate line. That works fine, and yeah, I like that better.

1.0/Dockerfile
```diff
@@ -40,6 +40,8 @@ RUN yum install -y centos-release-dotnet centos-release-scl-rh && \
       $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all -y && \
+    # yum cache files may still exist (and quite large in size)
```
tmds: nit: removing whitespace at the front makes the comments more visible (and allows grouping using comments)

PR #182, opened 2018-05-28
Use rh-dotnet21 scl

tmds:


tmds:
I have included replacing rh-nodejs6 with rh-nodejs8 in this PR as a separate commit to avoid merge conflicts.

PR #186, opened 2018-05-30
Sync with downstream containers.

aslicerh:
This also updates the tests to the latest version checks.

### DIFF
--- a/2.0/runtime/test/testcommon
+++ b/2.0/runtime/test/testcommon
@@ -2,7 +2,7 @@
 # The code is duplicated in both folders because the images are built
 # and tested after copying the sub-directories into a container repo.
 
-test_port=8080
+TEST_PORT=${TEST_PORT:-8080}
 CURLE_COULNDT_CONNECT=7
 
 if [ "$DEBUG" != "" ]; then
@@ -83,7 +83,7 @@ docker_run_as_d() {
   local user="$2"
 
   local output;
-  output=$(docker run --user="${user}" -d -p ${test_port}:8080 ${image} 2>&1)
+  output=$(docker run --user="${user}" -d -p ${TEST_PORT}:8080 ${image} 2>&1)
   if [ $? -ne 0 ]; then
     error "$output"
     echo "no-container"
@@ -95,7 +95,7 @@ docker_run_as_d() {
 docker_run_d() {
   local image="$1"
   local output;
-  output=$(docker run -d -p ${test_port}:8080 ${image} 2>&1)
+  output=$(docker run -d -p ${TEST_PORT}:8080 ${image} 2>&1)
   if [ $? -ne 0 ]; then
     error "$output"
     echo "no-container"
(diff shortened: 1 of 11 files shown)


aslicerh:
[cibuild]

aslicerh:
[cibuild]

aslicerh:
[cibuild]

aslicerh:
[cibuild]

aslicerh:
[cibuild]

aslicerh:
[cibuild]

aslicerh:
[cibuild]

2.0/runtime/test/testcommon
```diff
@@ -191,6 +191,6 @@ container_url() {
     error "$output"
     echo "no-url"
   else
-    echo "http://$output:${test_port}"
+    echo "http://$output:8080"
```
tmds: `${TEST_PORT}`?

tmds:
Can you add TEST_PORT to the overview in build.sh?

https://github.com/redhat-developer/s2i-dotnetcore/blob/70273cc382a5641ab1a1d09d09d7a89a64513c7d/build.sh#L7-L17

And make similar changes to the testcommon files in 2.1-dev branch, after this PR is done.

2.0/runtime/test/testcommon
aslicerh: No, apparently that is the actual port of the container, not the published port. They were just both '8080' before.

aslicerh:
@tmds 
Yeah, I'll add TEST_PORT to the overview, and I already planned to make the change for 2.1 dev. I haven't yet because part of me putting this out was to test the PR bot.

2.0/runtime/test/testcommon
tmds: Doesn't `curl` use this url on the host to connect to the container?

tmds: I'm surprised... but if it works... it works.

aslicerh: Surprised that it used the docker virtual network, or surprised that it was 8080 and not TEST_PORT on the docker virtual network?

tmds: The second.

aslicerh: I think it makes sense. Docker has to be able to talk to this stuff somehow when you have an EXPOSE directive. So it has an actual virtual network interface for that instead of just hiding it in an internal implementation. If you want to talk to said container over a normal interface then you have to use the bound port.

tmds: I hadn't given the docker virtual network much thought. I was thinking about how namespaces make the container port invisible to the host. Docker does more than namespaces, such as the virtual network. And then some stuff becomes visible again. Interesting!

tmds:
Don't forget the overview in build.sh.

aslicerh:
[cibuild]

PR #200, opened 2018-06-25
Trim nodejs dependencies

tmds:
This reduces image size and avoids having to respin the images on kernel CVEs.


tmds:
Thanks for reviewing @omajid @aslicerh !

PR #205, opened 2018-07-02
2.1: Add DOTNET_RM_SRC to clear src folder at the end of assemble

tmds:


tmds:

PR #212, opened 2018-08-20
2.1: cleanup: remove tools dir from PATH

tmds:
This is part of the dotnet scl enable script.

### DIFF
--- a/2.1/build/Dockerfile
+++ b/2.1/build/Dockerfile
@@ -2,7 +2,7 @@ FROM dotnet/dotnet-21-runtime-centos7
 # This image provides a .NET Core 2.1 environment you can use to run your .NET
 # applications.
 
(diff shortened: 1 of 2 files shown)


tmds:
Thanks for reviewing @aslicerh .

CI is failing with `Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock`. Maybe Docker is dead on the CI server.

aslicerh:
@tmds Yeah, I noticed that last week. The docker service is running, but the user that the bot runs as doesn't seem to have permission to talk with it now (for some reason). I added the user to the docker group last week, but it looks like I may need to get it to initiate a new session so that the group changes take effect.

aslicerh:
[cibuild]

tmds:
[cibuild]

test has timed out

Issue #213, opened 2018-08-20
Allow DOTNET_SSL_DIRS to be volumes mounted in the application container


tmds:
Currently DOTNET_SSL_DIRS are 'resolved' during assemble, and then used as-is during run.

It would be nice if DOTNET_SSL_DIRS are resolved each time. This allows the certificates to come from places that are not available at assemble, such as volumes which are mounted into the application container.

tmds:
Implemented in https://github.com/redhat-developer/s2i-dotnetcore/pull/214

PR #214, opened 2018-08-21
Allow DOTNET_SSL_DIRS to be volumes mounted into the application cont…

tmds:
This enables DOTNET_SSL_DIRS to include volumes that are mounted into the container.
Before this changes, only directories that were part of the build container ould be used as DOTNET_SSL_DIRS.

Implements https://github.com/redhat-developer/s2i-dotnetcore/issues/213.

### DIFF
--- a/2.1/runtime/README.md
+++ b/2.1/runtime/README.md
@@ -76,11 +76,11 @@ They must not to be overridden.
 
     These variables contain the working directory (`/opt/app-root/app`) and default CMD of the runtime image (`default-cmd.sh`).
 
-* **DOTNET_SSL_CERT_DIR**
+* **DOTNET_SSL_DIRS**
 
-    This variable is set to `/opt/app-root/ssl_dir`. This directory is not present in the runtime image.
-    When it added (e.g. by s2i assemble), it is used as `SSL_CERT_DIR`. `SSL_CERT_DIR` is used by applications (including `dotnet`)
-    to load CA certificates.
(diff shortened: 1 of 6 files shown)


2.1/runtime/README.md
```diff
@@ -76,11 +76,11 @@ They must not to be overridden.
...
     These variables contain the working directory (`/opt/app-root/app`) and default CMD of the runtime image (`default-cmd.sh`).
 
-* **DOTNET_SSL_CERT_DIR**
+* **DOTNET_SSL_DIRS**
 
-    This variable is set to `/opt/app-root/ssl_dir`. This directory is not present in the runtime image.
-    When it added (e.g. by s2i assemble), it is used as `SSL_CERT_DIR`. `SSL_CERT_DIR` is used by applications (including `dotnet`)
-    to load CA certificates.
+    Used to specify a list of folders with additional certificates to trust. The certificates are trusted by each process that runs
+    during the build and all processes that runs in the image after the build (including the application that was built). The folders
```
aslicerh: Run here should be singular.

aslicerh:
As you said, I'm not worried about these changes breaking undocumented behavior. That said, if a user needed to add certs for use at runtime, but wanted to rm src, is there a way to do that currently?

tmds:
> That said, if a user needed to add certs for use at runtime, but wanted to rm src, is there a way to do that currently?

The user should publish the certificates with his application and include the published certificate directory in DOTNET_SSL_DIRS.

tmds:

PR #215, opened 2018-08-23
Make older sdks aware of latest ASP.NET Core patch version

tmds:


tmds:
CI failed and server is unreachable.

The test passed on my machine.

tmds:
Thanks @omajid for reviewing!

tmds:
@aslicerh it should have passed. The test is running against the released sdks and they will now pick up the latest asp.net core version.

tmds:
Ah I see, you mean the released rpms already contain a version that isn't known to the tests.

While we're chatting and since I haven't seen downstream changes. Did you include the 3xx sdk package in the Dockerfiles explicitly? And extend the sdk list to include the 401 version?

tmds:
> Ah I see, you mean the released rpms already contain a version that isn't known to the tests.

I ran the tests against CentOS, thats why they passed on my machine.

aslicerh:
Yeah, that's what I mean. And with the changes to the dockerfile and test/run, everything passes on RHEL.

This would have been the first time we had side-by-side SDK packages, and no, I did not install them side-by-side. It wasn't until you brought up the asp.net issue today that I realized we need to start adding new install lines to the yum list in the dockerfile to side-by-side the SDKs. I'm going to do that on this new respin though.

tmds:
Thanks @aslicerh !

Issue #222, opened 2018-09-10
Support building from pre-compiled applications


tmds:
The s2i builder image should support building a deployable image from a pre-compiled application.

The java s2i builder supports this feature, see [Building and Deploying from Binary Artifacts
](https://access.redhat.com/documentation/en-us/openshift_online/3/html/using_images/source-to-image-s2i#s2i-images-java-deploy-applications-from-binary)

PR #225, opened 2018-09-13
Support building an image from a pre-compiled application

tmds:
Fixes https://github.com/redhat-developer/s2i-dotnetcore/issues/222


2.1/build/s2i/bin/assemble
```diff
@@ -17,77 +17,153 @@ if [ -d /tmp/src ]; then
...
+      # single project file
+      DOTNET_STARTUP_PROJECT="${DOTNET_STARTUP_PROJECT:-.}"
+      if [ -n "$DOTNET_ASSEMBLY_NAME" ]; then
+        APP_DLL_NAME="${DOTNET_ASSEMBLY_NAME}.dll"
+      else
+        PROJFILE=$(basename ${PROJFILES[0]})
+        APP_DLL_NAME="${PROJFILE: : -7}.dll"
+      fi
+    else
+      # zero or more project files -> fail
```
aslicerh: zero, or mare than 1 project files -> fail

2.1/build/s2i/bin/assemble
```diff
@@ -17,77 +17,153 @@ if [ -d /tmp/src ]; then
...
+      if [ ${#PROJFILES[@]} -eq 0 ]; then
+        echo "error: DOTNET_STARTUP_PROJECT has no project file"
+      else
+        echo "error: DOTNET_STARTUP_PROJECT contains multiple project files"
+      fi
+      echo "You can specify the startup project by adding an '.s2i/environment' file to the source repository."
+      # find all projects and strip leading './'
+      PROJFILES=(`find -name "*.??proj" | sed "s|^\./||"`)
```
aslicerh: On the line below this, you don't really need to check for 0 project files, because the error message is no more helpful than you've already shown on line 73. This means you can move this find and sed on line 79 inside the block starting on line 83 and only do it when necessary,

tmds: The earlier check is verifying `DOTNET_STARTUP_PROJECT, while this is trying to help the user figure out what projects are available in the source code repository (independent of DOTNET_STARTUP_PROJECT).

aslicerh: Ah, yes, I see. Ok, this is fine.

PR #228, opened 2018-09-19
Add DOTNET_RESTORE_DISABLE_PARALLEL to assemble

fbac:
If true, run the s2i build with --disable-parallel


kareiva:
Please backport this for 2.0 too.

aslicerh:
In addition to getting backported to 2.0, can something be added to the build/README.md, under Environment Variables, for this?

fbac:
Added envvar in 2.0 SDK, and also some docs.

2.1/build/s2i/bin/assemble
```diff
@@ -138,6 +138,11 @@ for SOURCE in $DOTNET_RESTORE_SOURCES; do
   RESTORE_OPTIONS="$RESTORE_OPTIONS --source $SOURCE"
 done
 
+# If true, run dotnet restore with --disable-parallel
+if [ -n "$DOTNET_RESTORE_DISABLE_PARALLEL" ]; then
```
tmds: Can you write this check as `== "true"`?

2.0/build/README.md
```diff
@@ -110,6 +110,11 @@ a `.s2i/environment` file inside your source code repository.
     Used to specify the list of NuGet package sources used during the restore operation. This overrides 
     all of the sources specified in the NuGet.config file.
 
+* **DOTNET_RESTORE_DISABLE_PARALLEL**
```
tmds: 2.0 is EOL in October. No need to add this to the 2.0 image.

2.1/build/README.md
```diff
@@ -110,6 +110,11 @@ a `.s2i/environment` file inside your source code repository.
     Used to specify the list of NuGet package sources used during the restore operation. This overrides 
     all of the sources specified in the NuGet.config file.
 
+* **DOTNET_RESTORE_DISABLE_PARALLEL**                                                                                                                                        
+                                                                                                                                                                             
+    Used to disable restoring multiple projects in parallel. Set this variable to true for it to take effect.
```
tmds: Suggested wording:

When set to `true` disables restoring multiple projects in parallel. This reduces restore timeout errors when the build container is running with low cpu limits. Defaults to `false`.

tmds:
See comments.

Please also add a test. You can set the variable here: https://github.com/redhat-developer/s2i-dotnetcore/blob/master/2.1/build/test/asp-net-hello-world-envvar/.s2i/environment and a check _something_ shows up in the build log similar to how we test the other envvars:
https://github.com/redhat-developer/s2i-dotnetcore/blob/dcaca684118ad1cd7df0b7e81a4182d43b84303a/2.1/build/test/run#L334-L357

tmds:
> Please backport this for 2.0 too.

@kareiva users should move to 2.1 at this point, 2.0 is EOL in October 1, we aren't pushing out new images.

fbac:
@tmds for the test I could add something like:
echo "Restore running with:" $RESTORE_OPTIONS

So the complete options shows up, and then check for "disable-parallel" in the test. With this also other options can be checked if added in the future.

Or just a simple static echo "Running with disable-parallel"

What would you like/think?

tmds:
Lets use the hello-world-envvar2 test project. If we increase the [DOTNET_VERBOSITY](https://github.com/redhat-developer/s2i-dotnetcore/blob/master/2.1/build/test/asp-net-hello-world-envvar2/.s2i/environment#L6) to `detailed`, we can verify the build log contains 'Running non-parallel restore`.

fbac:
Added test and vars to hello-world-envvar2.

tmds:
[cibuild]

tmds:
[cibuild]

2.1/build/test/run
```diff
@@ -340,6 +340,8 @@ test_config_1() {
   assert_contains "${s2i_build}" "/opt/app-root/src/src/app/bin/Debug/netcoreapp2.1/SampleApp.dll"
   # DOTNET_RESTORE_SOURCES=https://api.nuget.org/v3/index.json https://www.myget.org/F/s2i-dotnetcore
   assert_contains "${s2i_build}" "Installing S2iDotNetCoreDummy"
+  # DOTNET_RESTORE_DISABLE_PARALLEL=true
+  assert_contains "${s2i_build}" "Running non-parallel restore"
```
tmds: This check needs to be in test_config_2
This part of the test also needs an update because we changed DOTNET_VERBOSITY.
https://github.com/redhat-developer/s2i-dotnetcore/blob/4ef992a5f85234b665e398918b8581af90f5bc38/2.1/build/test/run#L384-L387

fbac: changed to test_config_2

But in test_config_2 there is also this check:
assert_contains "${s2i_build}" "DOTNET_VERBOSITY=normal"

Isn't this going to fail because we've changed in asp-net-hello-world-envvar2/.s2i/environment the verbosity to detailed?

tmds: Yes, that is the second part of my comment. You need to update =normal in the comment and check to =detailed.

fbac: Changed.

tmds:
[cibuild]

Systemic33:
How long time do you reckon it takes for this change to be available as an Openshift user?
ie. to show up here: https://access.redhat.com/containers/?tab=overview#/registry.access.redhat.com/dotnet/dotnet-21-rhel7

tmds:
guesstimate: first week of October

Systemic33:
Alright, good to know 👍

PR #240, opened 2018-11-08
Don't show retired image streams in OpenShift Console.

tmds:
This is done by adding the 'hidden' tag.


aslicerh:
Oh, good. I was kind of wondering what was going to happen with those.

Issue #231, opened 2018-10-01
runtime image: support building from pre-compiled applications


tmds:
https://github.com/redhat-developer/s2i-dotnetcore/issues/222 added support to the build image to build an application image from a pre-compiled application.

If we add this feature to the runtime image, the resulting application images will be smaller.
This means the runtime image becomes an s2i image where the _source_ is a pre-compiled application.

CC @aslicerh @omajid

PR #241, opened 2018-11-09
2.1: runtime image: support building from pre-compiled applications

tmds:
Implements https://github.com/redhat-developer/s2i-dotnetcore/issues/231


tmds:
Thanks @aslicerh ! I've pushed some additional commits that update the readme. Going to merge as is.

PR #242, opened 2018-11-10
Fixed assemble script to restore packages only once

eran-gil:
In the current assemble script the 'dotnet publish' does a restore in addition to the 'dotnet restore' command one line above it.
It caused me a lot of issues when I attempted to run a build on openshift for dotnet core 2.1.

### DIFF
--- a/2.1/build/s2i/bin/assemble
+++ b/2.1/build/s2i/bin/assemble
@@ -193,7 +193,7 @@ if [ "$BUILD_TYPE" == "source" ]; then
       echo "---> Restoring test project ($TEST_PROJECT) dependencies..."
       dotnet restore "$TEST_PROJECT" $RESTORE_OPTIONS $VERBOSITY_OPTION
       echo "---> Running test project: $TEST_PROJECT..."
-      dotnet test "$TEST_PROJECT" -f "$DOTNET_FRAMEWORK" $VERBOSITY_OPTION
+      dotnet test "$TEST_PROJECT" -f "$DOTNET_FRAMEWORK" $VERBOSITY_OPTION --no-restore
   done
 
   # publish application


tmds:
Can you be more specific about the issues this is solving for you?

If we add this to `dotnet publish` we also want to add it to `dotnet test`.
I wonder if adding this could break something that worked before.

erangil2:
Unfortunately, my project doesn't have a test project so I didn't think about it.
When using dotnet publish/test it also restores packages, which is done one line above it.
It caused me a problem because I use it in a network that is disconnected from the WWW, so I use a custom restore source by adding the 'DOTNET_RESTORE_SOURCES' parameter to the OpenShift build process. The restore completes perfectly, but the 'DOTNET_RESTORE_SOURCES' is not used in the publish command since the publish should just publish and not restore, and then it tries to restore anyways without the correct restore sources, and this fails the whole process for no good reason.

tmds:
I'd be interested to learn why the publish doesn't realize it can skip the restore and still feels the need to contact the nuget server. A build log with DOTNET_VERBOSITY set to diag may help uncover this. Could you send me one at tdeseyn AT redhat.com?

Can you update the PR to pass `--no-restore` to  `dotnet test` too?

erangil2:
Unfortunately I can't because the logs are in an isolated network where I can't export the logs outside to the WWW.
It might be due to the fact that some of the packages don't match the exact requirements, but find some "close-enough" alternatives, and the publish command does it again to attempt find a better solution.
However, I think it's a good practice to add the `--no-restore` parameter as it would save time from even checking the packages again after the restore command has done it already.
For the PR update, I've done it just now :)

tmds:
Can you take a look at the log? Or do you have an idea on how it can be reproduced?
If we can reproduce it, we can add it to our tests.

erangil2:
Tried finding a relevant log from OpenShift with no success, so please merge it as-is. Thanks

PR #245, opened 2018-11-26
2.2: runtime image: support building from pre-compiled applications

tmds:
Port https://github.com/redhat-developer/s2i-dotnetcore/pull/241 to 2.2


tmds:
Thank you for reviewing @aslicerh

Issue #249, opened 2018-12-05
DOTNET_SSL_DIRS doesn't work in a runtime image


nordseth:
When running using a runtime build with dotnet 2.1 the "DOTNET_SSL_DIRS" doesn't work, the container fails with
"rm: cannot remove '/opt/app-root/src': Is a directory"

The issue seems to be the following line in /contrib/etc/trust_ssl_dirs.
The /opt/app-root/src foldes does not exist

```
  # For backwards compatibility we need to run this in the src folder.
  # Create one if it does not exist.
  if [ ! -d /opt/app-root/src ]; then
    mkdir /opt/app-root/src
    _REMOVE_SRC_DIR=true
  fi

...

  # Remove temporary src folder.
  if [ "$_REMOVE_SRC_DIR" = true ] ; then
    rm /opt/app-root/src
  fi
```

tmds:
> rm /opt/app-root/src

This is missing an '-r' flag because it is a directory.

PR #250, opened 2018-12-05
Fix DOTNET_SSL_DIRS in runtime image

tmds:
For DOTNET_SSL_DIRS in the runtime image a temporary 'src' directory
is created. This folder was removed with 'rm' instead of 'rmdir'
which fails for directories.

Fixes https://github.com/redhat-developer/s2i-dotnetcore/issues/249

CC @nordseth


tmds:
Thanks @aslicerh 
@nordseth this will be part of the next security release.

PR #257, opened 2019-01-29
Update dotnet-pgsql-persistent.json

alberttwong:
removed readiness probe per https://github.com/redhat-developer/s2i-aspnet-musicstore-ex/issues/13


tmds:
Thanks!

PR #262, opened 2019-02-22
Supress 'ls' stderr for /tmp/artifacts

tmds:


tmds:
@aslicerh thanks for reviewing! Please include this in the images that have the incremental feature.

PR #270, opened 2019-04-22
Fix test_s2i_usage_projfile tests to handle indeterministic ordering of proj files.

aslicerh:
The order of the proj files as listed by the assemble script in s2i is not guaranteed to produce the same order each time, so the output is not static. This is fine, but the usage test needed to be changed to handle any order of the proj files.

### DIFF
--- a/2.1/build/test/run
+++ b/2.1/build/test/run
@@ -71,9 +71,9 @@ test_s2i_usage_projfile()
   local expected="error: DOTNET_STARTUP_PROJECT has no project file
 You can specify the startup project by adding an '.s2i/environment' file to the source repository.
 The source repository contains the following projects:
-- proj1/proj1.csproj
-- proj2/proj2.fsproj
-Update the '.s2i/environment' file to specify the project you want to publish, for example DOTNET_STARTUP_PROJECT=proj1/proj1.csproj."
+- proj\d/proj\d.csproj
+- proj\d/proj\d.fsproj
+Update the '.s2i/environment' file to specify the project you want to publish, for example DOTNET_STARTUP_PROJECT=proj\d/proj\d.csproj."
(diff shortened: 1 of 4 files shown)


tmds:
Lgtm, alternatively, you could sort in assemble to make it fixed order and alphabetical.

2.1/build/test/run
```diff
@@ -71,9 +71,9 @@ test_s2i_usage_projfile()
   local expected="error: DOTNET_STARTUP_PROJECT has no project file
 You can specify the startup project by adding an '.s2i/environment' file to the source repository.
 The source repository contains the following projects:
-- proj1/proj1.csproj
-- proj2/proj2.fsproj
-Update the '.s2i/environment' file to specify the project you want to publish, for example DOTNET_STARTUP_PROJECT=proj1/proj1.csproj."
+- proj\d/proj\d.csproj
+- proj\d/proj\d.fsproj
```
tmds: Actually, they are still in an order: one is a csproj, the other an fsproj.

aslicerh:
@tmds I don't really like the idea of changing the assemble script (something that has real function) just for the sake of the test. If the order being inconsistent was a functional issue, then I'd be more considerate of it.

tmds:
> just for the sake of the test

an ordered list is also nice UX.

aslicerh:
@tmds Do you have a preference between the two?

tmds:
I have a slight preference to order in assemble, because the user gets a nicer list then also.

aslicerh:
@tmds Can I get one more look over and approval for this, since it has been changed?

tmds:
LGTM. You can merge if the test suite passes. There seems to be no more CI happening here, a local test run is fine.

PR #282, opened 2019-07-14
remove versions that are EOL

gabemontero:
per request from @siamaksade from OpenShift PM, this PR removes EOF version from the imagestream file

@bparees - can you merge PRs in this repo, or do you know who we should bump to merge?  thanks

### DIFF
--- a/dotnet_imagestreams.json
+++ b/dotnet_imagestreams.json
@@ -80,69 +80,6 @@
                             "kind": "DockerImage",
                             "name": "registry.redhat.io/dotnet/dotnet-21-rhel7:2.1"
                         }
-                    },
-                    {
-                        "name": "2.0",
-                        "annotations": {
-                            "openshift.io/display-name": ".NET Core 2.0",
-                            "description": "RETIRED: Build and run .NET Core 2.0 applications on RHEL 7. For more information about using this builder image, including OpenShift considerations, see https://github.com/redhat-developer/s2i-dotnetcore/tree/master/2.0/build/README.md.",
-                            "iconClass": "icon-dotnet",
-                            "tags": "hidden,builder,.net,dotnet,dotnetcore,rh-dotnet20",
-                            "supports": "dotnet:2.0,dotnet",
-                            "sampleRepo": "https://github.com/redhat-developer/s2i-dotnetcore-ex.git",
-                            "sampleContextDir": "app",
-                            "sampleRef": "dotnetcore-2.0",
-                            "version": "2.0"
-                        },
-                        "referencePolicy": {
-                            "type": "Local"
-                        },
-                        "from": {
-                            "kind": "DockerImage",
-                            "name": "registry.redhat.io/dotnet/dotnet-20-rhel7:2.0"
-                        }
-                    },
-                    {
-                        "name": "1.1",
-                        "annotations": {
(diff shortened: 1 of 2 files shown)


aslicerh:
Hi @gabemontero,

We haven't removed those images on purpose, but they are marked as "RETIED" and tagged as hidden.

If this is not good enough, and we need to remove these from an OpenShift standpoint then we need to get approval from our own PM. Also, if this is the case, the 1.0 and 1.1 images should be removed as well, because they are also EOL.

gabemontero:
thanks for the quick response @aslicerh 

and yeah, we previously were using RETIRED for the various EOL image as well, but the ask from our PM was to actually remove the content.

So to get that approval, do you have a contact?  Or @siamaksade - do you know who is the PM for openshift/dotnet ?

Also @siamaksade - the 1.0/1.1 images were not marked as EOL in your document, so please confirm you are in agreement with @aslicerh 's comments there

aslicerh:
Hey @gabemontero.

Our PM is @Bob-Davis. I'll reach out to him on our side for the ok.

As for confirmation, we match Microsoft's support policy, which can be found [here](https://dotnet.microsoft.com/platform/support/policy/dotnet-core). 1.0 and 1.1 just went EOL recently, but we've already removed them from the redhat container catalog and marked them retired here. [Here's our own lifecycle docs as well](https://access.redhat.com/support/policy/updates/net-core).

Just out of curiosity, what is the expectation on timing for removing EOL'ed images from imagestreams moving forward?

gabemontero:
thanks for reaching out to @Bob-Davis and the support links @aslicerh 

so yeah, given the recent nature of the 1.0/1.1 EOL, I could see why @siamaksade might not have included it in the req document he sent me .... I'll update the PR once he confirms there was not some other unexpected reason one.

@siamaksade would also be the one to answer the questions around desired frequency of updates in this space

gabemontero:
ok realizing that @siamaksade might be travelling I just went ahead and removed 1.0/1.1 as well

at worst I'll see him face to face on Wednesday and can confirm he is on board at that time

siamaksade:
@gabemontero Super. I have updated the doc to include 1.0/1.1 removal as well

gabemontero:
@siamaksade reached out to @Bob-Davis and he is on PTO until July 29, so we'll have to stay on hold here and proceed with our current sprint feature without this change (where we handle this one via bugzilla if Bob approves).

aslicerh:
@gabemontero I already reached out to @Bob-Davis and got approval via our internal mailing list. So we're good to go on that front.

You need to update the PR though. The imagestream files define two imagestreams, one for runtime images and one for build/SDK images. You only removed build/sdk images, the 2.0 runtime image is still there. (There is only one image for 1.0 and 1.1, we didn;t start splitting the images until 2.0)

gabemontero:
ok in meetings now @aslicerh but will see about updating per your comment this evening europe time (4 or 5 hours perhaps if I get to it)

gabemontero:
imagestreams updated @aslicerh 

PTAL

aslicerh:
Looks good to me. Thanks @gabemontero

gabemontero:
cool thanks @aslicerh !!

PR #287, opened 2019-08-22
Fix condition in build.sh

aslicerh:
build.sh should use '||' not '-o'. Surprisingly, works correctly in 3 out of 4 use cases anyway. :/

-o should only be used inside of '[ ]'


tmds:
Thanks!

PR #289, opened 2019-09-03
Reduce log output size for DEBUG=y

tmds:
Most log output comes from DOTNET_VERBOSITY.
This changes the test_config_2 to run with verbosity 'normal' instead of 'detailed'.
The test for DOTNET_RESTORE_DISABLED_PARALLEL which needs 'detailed' verbosity is
moved into test_config_3 and filters the assemble output to what is needed by the test.


tmds:
This reduces `consoleText` on Jenkins from about 120MB to 15MB.

tmds:
I'm going to apply this to 2.1 and 2.2 after review.

tmds:
Passed on CI.

PR #290, opened 2019-09-05
Add DOTNET_PUBLISH_READYTORUN

tmds:
This enables using a new feature of the 3.0 sdk.

When set to `true`, the application will be compiled ahead-of-time. This reduces startup
time by reducing the amount of work the JIT needs to do when the application is loading.


tmds:
The preview9 sdk is needed to create framework-dependent R2R applications.

Except from the larger published output, I'm not aware of drawbacks. To be safe, I'm keeping this off by default.

PR #292, opened 2019-09-12
Add 3.0 images

tmds:


3.0/build/README.md
```diff
@@ -0,0 +1,202 @@
...
+    the JIT needs to do when the application is loading. Defaults to ``.
+
+* **DOTNET_RESTORE_SOURCES**
+
+    Used to specify the list of NuGet package sources used during the restore operation. This overrides 
+    all of the sources specified in the NuGet.config file. This variable cannot be combined with `DOTNET_RESTORE_CONFIGFILE`.
+
+* **DOTNET_RESTORE_CONFIGFILE**
+
+    Specifies a NuGet.Config file to be used for restore operations. When the file is not present, it won't cause the build to fail.
```
tmds: > When the file is not present, it won't cause the build to fail.

This is no longer accurate.

PR #304, opened 2019-11-28
Define fedora imagestreams

tmds:


tmds:
I've tested the file using `crc`.

PR #306, opened 2019-12-05
Update README, imagestreams, scripts for .NET Core 3.1

tmds:


tmds:
I verified this worked with the published images.

PR #311, opened 2020-01-07
Mark 2.2 as retired, remove it from builds, imagestreams, and doc examples.

aslicerh:


tmds:
Thanks!

cricket007:
May I ask why 2.1 is maintained but 2.2 is retired?

tmds:
We follow the upstream/Microsoft life cycle.
2.2 is no longer being patched for security issues.

This is the Red Hat life cycle page: https://access.redhat.com/support/policy/updates/net-core.
It matches with Microsoft life cycle: https://dotnet.microsoft.com/platform/support/policy/dotnet-core.

If you are on 2.2, it is recommended to move to 3.1. 3.1 is an LTS release that is supported till December 2022.

cricket007:
Thanks for the links! So my confusion starts at Microsoft. Of course 😅

PR #315, opened 2020-02-07
Update everything according to the new supported versions across all platforms.

aslicerh:


2.1/build/Dockerfile
```diff
@@ -47,8 +47,8 @@ ENV ENABLED_COLLECTIONS="$ENABLED_COLLECTIONS rh-nodejs10" \
 # Needed for the `dotnet watch` to detect changes in a container.
     DOTNET_USE_POLLING_FILE_WATCHER=true \
 # Make all sdks aware of the latest ASP.NET Core version
-    LatestPatchVersionForAspNetCoreApp2_1=2.1.7 \
-    LatestPatchVersionForAspNetCoreAll2_1=2.1.7
+    LatestPatchVersionForAspNetCoreApp2_1=2.1.14 \
+    LatestPatchVersionForAspNetCoreAll2_1=2.1.14
```
tmds: This is for making older SDKs aware of newer ASP.NET Core versions.
We can remove this, because we've dropped support for older sdks at some point.
But I don't mind if it sticks around either.

aslicerh: I'll leave it in for now and we can see about taking it out later.

PR #323, opened 2020-04-28
Remove LatestPatchVersionForAspNetCore* ENV vars and update expected versions for tests.

aslicerh:
These were added when there were multiple supported ASP.NET Core versions supported in a single container, which is no longer the case.

### DIFF
--- a/2.1/build/test/run
+++ b/2.1/build/test/run
@@ -43,25 +43,23 @@ OPENSHIFT_ONLY=${OPENSHIFT_ONLY:-false}
 test_dir="$(readlink -zf $(dirname "${BASH_SOURCE[0]}"))"
 source ${test_dir}/testcommon
 
-npm_version=6.4.1
+npm_version=6.13.4
 
 if [ "$IMAGE_OS" = "CENTOS" ]; then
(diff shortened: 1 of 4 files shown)


2.1/build/test/run
```diff
@@ -43,25 +43,23 @@ OPENSHIFT_ONLY=${OPENSHIFT_ONLY:-false}
...
-npm_version=6.9.0
 elif [ "$IMAGE_OS" = "RHEL8" ]; then
 # sdk version supported on RHEL8
-sdk_version=2.1.511
-aspnet_latest_app_version=2.1.15
-aspnet_latest_all_version=2.1.15
-npm_version=6.9.0
+sdk_version=2.1.513
+aspnet_latest_app_version=2.1.17
+aspnet_latest_all_version=2.1.17
```
tmds, line 57: Can we remove `aspnet_latest_all_version`/`aspnet_latest_app_version` and related tests also?
If we want to test this, let's do it early in `dotnet-regular-tests`.

aslicerh: Done.

tmds:
Thanks, Andrew!

PR #330, opened 2020-07-17
Enable use of 'dotnet' in Dockerfile RUN instructions on scl-based images.

tmds:
The 'dotnet' command cannot be used directly from Dockerfile RUN instructions
because the scl is not enabled.

This adds a script named 'dotnet' on PATH which enables the scl and invoke
'dotnet' in it. Once the scl is enabled, the 'dotnet' host takes preference
over the script because it is earlier on the PATH variable.

### DIFF
--- a/2.1/build/test/testcommon
+++ b/2.1/build/test/testcommon
@@ -41,15 +41,15 @@ fi
 
 _PODMAN_PATH=$(command -v podman)
 _DOCKER_PATH=$(command -v docker)
-if [ -f "$_DOCKER_PATH" ] && [[ "$(docker --version)" != *podman* ]]; then
-  info "Using docker"
-  USING_PODMAN=""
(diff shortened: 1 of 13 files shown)


2.1/build/test/testcommon
```diff
@@ -41,15 +41,15 @@ fi
 
 _PODMAN_PATH=$(command -v podman)
 _DOCKER_PATH=$(command -v docker)
-if [ -f "$_DOCKER_PATH" ] && [[ "$(docker --version)" != *podman* ]]; then
-  info "Using docker"
-  USING_PODMAN=""
-elif [ -f "$_PODMAN_PATH" ]; then
+if [ -f "$_PODMAN_PATH" ]; then
```
tmds, line 44: This is an unrelated change. `docker version` on my system no longer contains `podman`, and this check didn't work as expected. Now it simply prefers `podman` over `docker`.

PR #331, opened 2020-08-12
Restructure imagestreams

yselkowitz:
Following the pattern set by the SCLs, imagestreams are now described as
a combination of runtime and underlying OS versions. UBI 8 imagestreams
are offered alongside RHEL/CentOS 7 ones, with latest being moved to UBI 8,
and previously existing imagestream tags kept but hidden for compatibility.

### DIFF
--- a/dotnet_imagestreams_centos.json
+++ b/dotnet_imagestreams_centos.json
@@ -23,7 +23,7 @@
                         "name": "latest",
                         "annotations": {
                             "openshift.io/display-name": ".NET Core (Latest)",
-                            "description": "Build and run .NET Core applications on CentOS 7. For more information about using this builder image, including OpenShift considerations, see https://github.com/redhat-developer/s2i-dotnetcore/tree/master/3.1/build/README.md.\n\nWARNING: By selecting this tag, your application will automatically update to use the latest version of .NET Core available on OpenShift, including major versions updates.",
+                            "description": "Build and run .NET Core applications on UBI. For more information about using this builder image, including OpenShift considerations, see https://github.com/redhat-developer/s2i-dotnetcore/tree/master/3.1/build/README.md.\n\nWARNING: By selecting this tag, your application will automatically update to use the latest version of .NET Core available on OpenShift, including major versions updates.",
                             "iconClass": "icon-dotnet",
                             "tags": "builder,.net,dotnet,dotnetcore",
                             "supports": "dotnet",
@@ -33,13 +33,34 @@
                         },
                         "from": {
                             "kind": "ImageStreamTag",
-                            "name": "3.1"
+                            "name": "3.1-ubi8"
                         }
                     },
                     {
-                        "name": "3.1",
+                        "name": "3.1-ubi8",
                         "annotations": {
-                            "openshift.io/display-name": ".NET Core 3.1",
+                            "openshift.io/display-name": ".NET Core 3.1 (UBI 8)",
+                            "description": "Build and run .NET Core 3.1 applications on UBI 8. For more information about using this builder image, including OpenShift considerations, see https://github.com/redhat-developer/s2i-dotnetcore/tree/master/3.1/build/README.md.",
+                            "iconClass": "icon-dotnet",
+                            "tags": "builder,.net,dotnet,dotnetcore,dotnet31",
+                            "supports": "dotnet:3.1,dotnet",
+                            "sampleRepo": "https://github.com/redhat-developer/s2i-dotnetcore-ex.git",
+                            "sampleContextDir": "app",
+                            "sampleRef": "dotnetcore-3.1",
+                            "version": "3.1"
+                        },
+                        "referencePolicy": {
+                            "type": "Local"
+                        },
+                        "from": {
+                            "kind": "DockerImage",
+                            "name": "registry.access.redhat.com/ubi8/dotnet-31:3.1"
+                        }
+                    },
+                    {
+                        "name": "3.1-el7",
+                        "annotations": {
+                            "openshift.io/display-name": ".NET Core 3.1 (CentOS 7)",
                             "description": "Build and run .NET Core 3.1 applications on CentOS 7. For more information about using this builder image, including OpenShift considerations, see https://github.com/redhat-developer/s2i-dotnetcore/tree/master/3.1/build/README.md.",
                             "iconClass": "icon-dotnet",
                             "tags": "builder,.net,dotnet,dotnetcore,rh-dotnet31",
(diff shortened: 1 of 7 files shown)


tmds:
> Following the pattern set by the SCLs, imagestreams are now described as
a combination of runtime and underlying OS versions.

@yselkowitz can you provide some more context: is this pattern described somewhere? what other SCLs have adopted this? and what problem is this solving?

yselkowitz:
Once this filters down to openshift/library, it can then be pulled into openshift/library, then into the cluster-samples-operator, where it will go through image-ecosystem testing before landing there.

yselkowitz:
Any further concerns on this?

tmds:
I will test the updated files this week, and if everything works fine, I'll merge the PR.

dotnet_imagestreams_centos.json
```diff
@@ -97,7 +175,40 @@
...
+                            "tags": "runtime,.net-runtime,dotnet-runtime,dotnetcore-runtime",
+                            "supports": "dotnet-runtime",
+                            "version": "3.1"
+                        },
+                        "referencePolicy": {
+                            "type": "Local"
+                        },
+                        "from": {
+                            "kind": "DockerImage",
+                            "name": "registry.access.redhat.com/ubi8/dotnet-31-runtime:3.1"
```
tmds, line 196: The `_centos`  imagestreams are refering to the rhel ubi8 image. Is this intentional?

dotnet_imagestreams_centos.json
```diff
@@ -115,13 +226,46 @@
...
+                            "tags": "runtime,.net-runtime,dotnet-runtime,dotnetcore-runtime",
+                            "supports": "dotnet-runtime",
+                            "version": "2.1"
+                        },
+                        "referencePolicy": {
+                            "type": "Local"
+                        },
+                        "from": {
+                            "kind": "DockerImage",
+                            "name": "registry.access.redhat.com/ubi8/dotnet-21-runtime:2.1"
```
tmds, line 244: same here

templates/aspnet-2.x.json
```diff
@@ -14,14 +14,14 @@
             },
             "spec": {
                 "source": {
-                    "dockerfile": "FROM dotnet:2.1\nRUN bash -c 'dotnet new web -o /tmp/tmpfolder && rm -rf /tmp/tmpfolder && fix-permissions /opt/app-root'"
+                    "dockerfile": "FROM dotnet:2.1-ubi8\nRUN bash -c 'dotnet new web -o /tmp/tmpfolder && rm -rf /tmp/tmpfolder && fix-permissions /opt/app-root'"
```
tmds, line 17: I'm not a fan of how the suffix starts to show up here in templates and examples. For a user this is noise.

Most templates/examples work independent of the specific base used, though this changes makes them require a specific base.

dotnet_imagestreams_centos.json
yselkowitz: The SCL team does not plan on building any CentOS 8 images for those components which are now in UBI, as there is no practical need (UBI being freely available without a subscription).  Unless you have such images, using UBI across the board keeps OKD and OCP even with each other.

dotnet_imagestreams_centos.json
yselkowitz: Please see my response above.

templates/aspnet-2.x.json
yselkowitz: It was decided that the original tags (e.g. '2.1') could not simply be repurposed for UBI/RHEL 8 images, as this would trigger rebuilds of users' existing applications and potentially introduce incompatibilities due to sometimes subtle differences between RHEL versions.  (Except for 'latest', which is anyways a floating tag, and the risk of incompatibilities between runtime versions is generally much more than between underlying userspace.)

Therefore, this new naming scheme was devised to give users the choice between compatibility and innovation.  If you have further questions about it, please feel free to contact me directly.

tmds:
We refer to these files in our documentation, like [installing image streams](https://access.redhat.com/documentation/en-us/net_core/2.1/html/getting_started_guide_for_rhel_8/gs_dotnet_on_openshift#install_imagestreams) and [sample applications](https://access.redhat.com/documentation/en-us/net_core/2.1/html/getting_started_guide_for_rhel_8/gs_dotnet_on_openshift#sample_apps).

Can we make these changes in the interest of backwards compatibility:

- keep the `dotnet_imagestreams_rhel8.json` file
- undo changes to templates. (Note that we are no longer updating these templates and will EOL them with .NET 2.1.)
- continue to accept `rhel7` and `rhel8` and `centos7` as arguments to the `install-imagestreams` scripts. (`rhel7` can be an alias for `rhel`).

yselkowitz:
The point of these changes is to alleviate the need for special steps to use UBI8 imagestreams, and to provide a consistent set of imagestreams across both OKD and OCP, and compatibility during upgrades.  To that end, could the documentation be updated to match instead of the other way around?

The templates currently reference obsolete imagestream tags (dotnet:3.0 and postgresql:9.5) which are no longer provided, so they require some form of update regardless.  How would you prefer to handle this?

cc: @gabemontero

tmds:
> The point of these changes is to alleviate the need for special steps to use UBI8 imagestreams, and to provide a consistent set of imagestreams across both OKD and OCP, and compatibility during upgrades.

This happens through the changes made to `dotnet_imagestreams.json`, which are backwards-compatible.

OpenShift doesn't always provide the latest version of .NET Core, and it's not uncommon for a user to manually update their image streams. I want to avoid breaking these users, which may have some automation set up around files other than `dotnet_imagestreams.json`.

> The templates currently reference obsolete imagestream tags (dotnet:3.0 and postgresql:9.5) which are no longer provided, so they require some form of update regardless. How would you prefer to handle this?

We're no longer updating any templates. We let them go EOL with their respective .NET Core version. We don't have any templates for the latest version of .NET Core.

yselkowitz:
Then would it suffice to make `dotnet_imagestreams_rhel8.json` a symlink to `dotnet_imagestreams.json`, and rhel7 and rhel8 aliases to rhel in the install scripts?  We really want to avoid fracturing imagestream tags.

If `dotnet-examples.json` is not being updated from 3.0 to 3.1, it would have to be removed from openshift.  As for `dotnet-pgsql-persistent.json`, 2.1 is still supported for almost another year, but `postgresql:9.5` is out of support and no longer provided.  Do you want that updated to 9.6 so that it can still be provided?

tmds:
I think your main interest is consuming `dotnet_imagestreams.json` and `dotnet_imagestreams_centos.json`.

I want to avoid breaking users of `dotnet_imagestreams_rhel8.json`, people that use `install-imagestreams` with `rhel7` or `rhel8` arguments, and users of the examples/templates. I don't see a strong reason to break these users and I think we should keep this backwards compatible.

> If dotnet-examples.json is not being updated from 3.0 to 3.1, it would have to be removed from openshift.

Yes, it should be removed from OpenShift.

> As for dotnet-pgsql-persistent.json, 2.1 is still supported for almost another year, but postgresql:9.5 is out of support and no longer provided. Do you want that updated to 9.6 so that it can still be provided?

This one may be removed from OpenShift as well. We should still update the template so it uses a supported version of postgresql. Our documentation refers to this template.

yselkowitz:
I've revised this PR, if you could please review again.

tmds:
@yselkowitz This breaking users of `dotnet_imagestreams_rhel8.json` and users passing `rhel8` as an argument to the `install-imagestreams` scripts? Can you revert those changes, or elaborate why this is important?

yselkowitz:
Repushed with discussed changes.

yselkowitz:
Removal of deprecated templates from OpenShift filed: https://github.com/openshift/library/pull/250

install-imagestreams.ps1
```diff
@@ -203,23 +203,24 @@ if($Remove.IsPresent)
 }
 
 # The OS parameter is required.
-switch ( $OS )
+switch -wildcard ( $OS )
```
tmds, line 206: You are strong in the art of powershell :smile:

tmds (approved):

tmds:
@yselkowitz thank you for putting us in line with the rest of the image streams in OpenShift, and taking care of our deprecated templates!

PR #333, opened 2020-08-26
scl_enable_dotnet: minor implementation improvements

tmds:
Source existing scl_enable script and exec dotnet directly.


tmds:
Thanks for reviewing @omajid !

PR #334, opened 2020-09-03
Have dockerfile-run tests create their needed folders.

aslicerh:
This test currently does not create its own folder, nor is that folder (being normally empty) an existing part of the file tree.

Because the test also creates a dockerfile, this has a fun side effect of overwriting the dockerfile in the base folder for the container if you run the tests from there.

This should prevent that from happening.


tmds:
Thanks for fixing!

PR #339, opened 2020-10-08
build.sh: when IMAGE_OS is set to empty string use host OS to determine base image

tmds:


tmds:
Andrew, ptal and merge if it is ok for you.

aslicerh (approved):

PR #341, opened 2020-11-30
Update imagestreams

tmds:
- Add 5.0 images.

- Retire _centos, _fedora, _rhel8 in favor of dotnet_imagestreams ubi8 images.

- Drop .git from s2i-dotnetcore-ex suffix because it breaks OpenShift Console 'Edit Source Code' url
(for example: https://github.com/redhat-developer/s2i-dotnetcore-ex.git/tree/dotnet-5.0).

### DIFF
--- a/dotnet_imagestreams.json
+++ b/dotnet_imagestreams.json
@@ -4,7 +4,7 @@
     "metadata": {
         "name": "dotnet-image-streams",
         "annotations": {
-            "description": "ImageStream definitions for .NET Core on RHEL 7"
+            "description": "ImageStream definitions for .NET"
         }
     },
     "items": [
@@ -14,7 +14,7 @@
             "metadata": {
                 "name": "dotnet",
                 "annotations": {
-                    "openshift.io/display-name": ".NET Core"
+                    "openshift.io/display-name": ".NET"
                 }
             },
             "spec": {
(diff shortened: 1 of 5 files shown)


dotnet_imagestreams.json
```diff
@@ -193,7 +235,43 @@
...
                         "from": {
                             "kind": "ImageStreamTag",
-                            "name": "3.1-ubi8"
+                            "name": "5.0-ubi8"
+                        }
+                    },
+                    {
+                        "name": "5.0-ubi8",
+                        "annotations": {
+                            "openshift.io/display-name": ".NET 5.0 Runtime (UBI 8)",
```
aslicerh, line 244: I thought we were dropping the `.0` in `5.0` in display strings?

dotnet_imagestreams_centos.json
```diff
@@ -4,7 +4,7 @@
     "metadata": {
         "name": "dotnet-image-streams",
         "annotations": {
-            "description": "ImageStream definitions for .NET Core on CentOS"
+            "description": "ImageStream definitions for .NET Core on CentOS (Retired)"
```
tmds, line 7: Probably this gets used by OKD and we can't deprecate it. I'll update it instead.

dotnet_imagestreams.json
tmds: I'll update the display-name and descriptions to remove the .0.

dotnet_imagestreams.json
```diff
@@ -193,7 +235,43 @@
...
+                            "description": "Run .NET applications on UBI 8. For more information about using this image, including OpenShift considerations, see https://github.com/redhat-developer/s2i-dotnetcore/tree/master/5.0/runtime/README.md.",
```
tmds, line 245: I got this description by copy-paste. I see that some include a version number and others don't. I'll make it consistent by including a version number in all of them.

dotnet_imagestreams_rhel8.json
```diff
@@ -4,7 +4,7 @@
     "metadata": {
         "name": "dotnet-image-streams",
         "annotations": {
-            "description": "ImageStream definitions for .NET Core on RHEL 8"
```
aslicerh, line 7: This kind of implies to me that we don't have imagestreams that work on RHEL8, but this isn't quite true. The UBI imagestreams will use your RHEL8 subscriptions to install packages if they are setup on the host machine (from what I understand).

I don't know if there is a good way to *convey* that here, but I thought I'd bring it up.

tmds: I don't have an alternative in mind. I think Retired will make people look what they should use instead.

PR #347, opened 2021-01-26
install-imagestreams.ps1: fix help always being printed.

tmds:
Check the number of bound parameters instead of unbound arguments.

Fixes https://github.com/redhat-developer/s2i-dotnetcore/issues/346.

cc @faysou @omajid


aslicerh (approved):
LGTM

omajid (approved):
Can't test this, but it matches what stackoverflow says :shipit:

PR #351, opened 2021-02-11
Work around a regression in podman when using 'podman inspect -f "json ...'

aslicerh:


2.1/build/test/testcommon
```diff
@@ -176,7 +176,11 @@ docker_get_env() {
 
 docker_get_exposedports() {
   local image="$1"
-  docker inspect -f '{{json .Config.ExposedPorts}}' ${image}
+  if [ "$USING_PODMAN" ]; then
```
tmds, line 179: If the `jq` stuff works for `docker` too, I'd prefer to not split into separate case, and take the `jq` dependency.

tmds:
Thanks for fixing. You can merge when the tests pass for you.
If you want, you can run it on the CI server by changing the job configuration and point it to your `--repo` and `--ref`.

tmds (approved):

2.1/build/test/testcommon
aslicerh: I figured we'd just pull this out (and ditch the jq dependency) when the regression is fixed.

But I went ahead and changed the command to work for both. Had to change it slightly, because docker actually reacts slightly differently than podman to that command.

aslicerh:
@tmds I've updated the code and the CI job, but the test machines don't have jq installed. It _should_ work though. (Famous last words)

PR #365, opened 2021-05-27
Change how container-tool cleans up TMPDIR

aslicerh:


3.1/build/s2i/bin/container-tool
```diff
@@ -3,13 +3,15 @@
...
+BUILD_DIR="/tmp/container-tool"
 
 # command that builds the tool, used by the image build.
 if [ "$1" == "build-tool" ]; then
   set -x
 
-  mkdir /tmp/container-tool
-  pushd /tmp/container-tool
+  mkdir $BUILD_DIR
+  export TMPDIR=${BUILD_DIR}
```
tmds, line 13: Can you make `TMPDIR` something other than `BUILD_DIR`, so tmp files don't end up next to the source and csproj file? Like: `export TMPDIR=/tmp/dotnet-tmp`.
You'll need to add it to the `mkdir` and `rm` next to `BUILD_DIR`.

3.1/build/s2i/bin/container-tool
```diff
@@ -40,7 +42,9 @@ EOF
   popd
 
   set +e
-  rm -rf /tmp/* /tmp/.*
+  # /tmp/.dotnet/shm is created even with the TMPDIR change,
+  # so make sure to remove that as well.
```
tmds, line 46: You should be able to remove the `set +e` that is just above. It was meant for the wildcard matching.

3.1/build/s2i/bin/container-tool
```diff
@@ -3,13 +3,15 @@
...
 
 # command that builds the tool, used by the image build.
 if [ "$1" == "build-tool" ]; then
   set -x
 
-  mkdir /tmp/container-tool
-  pushd /tmp/container-tool
+  mkdir $BUILD_DIR
+  export TMPDIR=${BUILD_DIR}
+  pushd $BUILD_DIR
```
tmds, line 14: For consistency, can you use `"$BUILD_DIR"`?

tmds:
@aslicerh thank you for fixing this!

3.1/build/s2i/bin/container-tool
aslicerh: Done.

3.1/build/s2i/bin/container-tool
aslicerh: Done.

3.1/build/s2i/bin/container-tool
aslicerh: Kind of done. Let me know if you'd prefer multiple mkdirs and all instead of `mkdir -p` and a single full path.

3.1/build/s2i/bin/container-tool
```diff
@@ -3,13 +3,16 @@
 set -e
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+TMP_DIR="/tmp/dotnet-tmp"
+BUILD_DIR="${TMP_DIR}/container-tool"
```
tmds, line 7: Can you move these under the `if`, and squash `TMP_DIR` and `TMPDIR`?
```
export TMPDIR="/tmp/dotnet-tmp"
BUILD_DIR="${TMPDIR}/container-tool"
mkdir -p "$BUILD_DIR"
```

aslicerh: Personally, I usually prefer variables used across the script like this at the top of the script, instead of down somewhere else, though I guess since these are only used inside the if they could go in there.

Any reason not to move SCRIPT_DIR too, then?

tmds: The `TMP_DIR`/`TMPDIR` stuff is specific to the `build-tool` command, and the `SCRIPT_DIR` gets used also when we do `exec dotnet`.

tmds (approved):

5.0/build/test/run
```diff
@@ -678,12 +678,19 @@ test_incremental_without_packages()
   docker_rmi ${image}
 }
 
+test_repos() {
+    test_start
+
+    local output=$(docker_run_as $IMAGE_NAME 0 yum install -y zlib-devel)
```
tmds, line 684: You need to add some assert here, otherwise the test can't fail.

3.1/build/test/run
```diff
@@ -687,12 +686,19 @@ test_incremental_without_packages()
   docker_rmi ${image}
 }
 
+test_repos() {
+    test_start
+
+    assert_contains "$(docker_run_as $IMAGE_NAME 0 yum install -y zlib-devel)" "Complete!"
```
tmds, line 692: style nit: can you put the `docker_run_as` and the `assert` on different lines? It's consistent with the other tests, and applies the AAA (Arrange-Act-Assert) pattern.

tmds (approved):

PR #376, opened 2021-09-24
Misc. Container Updates

aslicerh:
The commit titles are pretty on point. This:
##### Created some .gitignore files
The tests tend to create some empty dirs that sit around. They're harmless, but I like not seeing them in the `git status` reports.
##### Specified nodejS module for RHEL8 containers
Something like this is necessary to change the module/stream used in rhel 8 to get newer versions of nodeJS. As a bonus, it makes it more obvious which nodeJS version we're using.
##### Changed DOTNET TOOL tests for how new npm versions work
When running tests for some beta images, I noticed that these tests failed because the new version of npm no longer spits out the package name that was installed after it is done. I fixed this by doing an `npm ls` call to get the list of npm packages, and seeing if our installed package was listed there. Unfortunately, `npm ls` throws an error (about 'extraneous' packages?), for no real reason that I can understand. It's seriously weird.

### DIFF
--- a/3.1/build/.gitignore
+++ b/3.1/build/.gitignore
@@ -1,0 +1,1 @@
+test/split_build
--- a/3.1/build/Dockerfile.rhel8
+++ b/3.1/build/Dockerfile.rhel8
@@ -3,7 +3,8 @@ FROM ubi8/dotnet-31-runtime
 # applications.
 
 ENV PATH=/opt/app-root/src/.local/bin:/opt/app-root/src/bin:/opt/app-root/node_modules/.bin:/opt/app-root/.dotnet/tools/:${PATH} \
-    STI_SCRIPTS_PATH=/usr/libexec/s2i
+    STI_SCRIPTS_PATH=/usr/libexec/s2i \
+    NODEJS_VERSION=14
 
 LABEL io.k8s.description="Platform for building and running .NET Core 3.1 applications" \
       io.openshift.tags="builder,.net,dotnet,dotnetcore,dotnet-31"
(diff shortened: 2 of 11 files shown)


3.1/build/Dockerfile.rhel8
```diff
@@ -3,7 +3,8 @@ FROM ubi8/dotnet-31-runtime
 # applications.
 
 ENV PATH=/opt/app-root/src/.local/bin:/opt/app-root/src/bin:/opt/app-root/node_modules/.bin:/opt/app-root/.dotnet/tools/:${PATH} \
-    STI_SCRIPTS_PATH=/usr/libexec/s2i
+    STI_SCRIPTS_PATH=/usr/libexec/s2i \
+    NODEJS_VERSION=14
```
tmds, line 7: I thought about adding this envvar when doing initial 6.0 work. I did not add it because it gets only used in one place, and an end-user has little/no use for it. I included the number directly in the `yum` command.

https://github.com/redhat-developer/s2i-dotnetcore/blob/f92d394bb526091d7f5bd9743634ff53609068da/6.0/build/Dockerfile.rhel8#L36

5.0/build/Dockerfile.rhel8
```diff
@@ -3,7 +3,8 @@ FROM ubi8/dotnet-50-runtime
 # applications.
 
 ENV PATH=/opt/app-root/src/.local/bin:/opt/app-root/src/bin:/opt/app-root/node_modules/.bin:/opt/app-root/.dotnet/tools/:${PATH} \
-    STI_SCRIPTS_PATH=/usr/libexec/s2i
+    STI_SCRIPTS_PATH=/usr/libexec/s2i \
+    NODEJS_VERSION=14
```
tmds, line 7: Same comment about the envvar.

3.1/build/.gitignore
```diff
@@ -0,0 +1 @@
+test/split_build
```
tmds, line 1: :+1:

Thanks for adding these `.gitignore` files.

tmds (approved):

tmds:
@aslicerh this is good to merge. Thanks!

PR #378, opened 2021-10-07
Remove content_sets.* files, as they are not needed to build these containers.

aslicerh:


tmds:
lgtm.

Unless it is important this goes on the main branch now, can you target the `dev_60` branch and remove the content_sets files from `6.0` as well?

Anything on `dev_60` will be merged as separate commits.

PR #381, opened 2021-10-14
Remove nodejs

aslicerh:
This removes nodejs from the 6.0 build image.

I also tried to replace the DEV_MODE behavior using some other, lighter weight packages.

I picked pyinotify because it has a small footprint, is available in the UBI packages (and is actually already installed in the base ubi base image). flock is used to keep it from launching a bunch of processes that will fail in the middle of the build process. dotnet watch is used so that the app will be killed and restarted on a file change.


tmds:
I think @omajid once said that maybe we could remove it. And my feedback was that I'd prefer to keep it if .NET 6 still had templates using `nodejs`. The `angular` template that you're removing from the tests in this PR is using it.

I don't think it was discussed further, or that someone decided we'd remove it?

PR #385, opened 2021-10-27
Reduce image size by not installing weak dependencies.

tmds:
For the runtime image this removes the cmake and some perl-* dependencies of nss_wrapper.
Neither are needed.
The installed size is reduced by 36M.

For the sdk image this removes the nodejs-docs and nodejs-full-i18n depedencies of nodejs.
The docs are not needed to build applications, and the internationalization is not needed
to build JavaScript front-ends.
The installed size is reduced by 68M.


tmds:
@aslicerh @omajid ptal.

aslicerh (approved):
LGTM

tmds:
Thanks for taking a look, Andrew.

PR #387, opened 2021-11-03
Remove DEV_MODE feature because it is no longer used by IDEs.

tmds:
Fixes https://github.com/redhat-developer/s2i-dotnetcore/issues/386.


6.0/build/Dockerfile.fedora
```diff
@@ -18,18 +18,14 @@ LABEL name="fedora/dotnet-60" \
...
-
 # Switch to root for package installs
 USER 0
 
 # Copy the S2I scripts from the specific language image to $STI_SCRIPTS_PATH.
 COPY ./s2i/bin/ /usr/libexec/s2i
 
 # TODO: use .NET package (dotnet-sdk-6.0).
-RUN INSTALL_PKGS="npm nodejs-nodemon rsync procps-ng findutils" && \
+RUN INSTALL_PKGS="npm procps-ng" && \
```
tmds, line 28: `nodejs-nodemon rsync` were added to support `DEV_MODE`.
I've removed `findutils` here, because the base image installs it also.

aslicerh (approved):
Looks fine to me.

PR #388, opened 2021-11-04
Remove unzip dependency which is no longer needed for installing vsdbg.

tmds:


6.0/build/Dockerfile.fedora
```diff
@@ -25,6 +25,10 @@ USER 0
 COPY ./s2i/bin/ /usr/libexec/s2i
 
 # TODO: use .NET package (dotnet-sdk-6.0).
+# Install packages:
```
tmds, line 28: I've added some comments about why we're installing different packages.

6.0/runtime/Dockerfile.fedora
```diff
@@ -46,13 +46,17 @@ COPY ./root/usr/bin /usr/bin
 # Copy the S2I scripts from the specific language image to $STI_SCRIPTS_PATH.
 COPY ./s2i/bin/ /usr/libexec/s2i
 
-# TODO: use .NET package (aspnetcore-runtime-6.0).
+# TODO: use .NET package (add: aspnetcore-runtime-6.0, remove:libicu).
```
tmds, line 49: `libicu` is there for the Microsoft binaries. It can be removed when we switch to our packages because our runtime package has a dependency.

6.0/runtime/Dockerfile.rhel8
```diff
@@ -46,21 +46,23 @@ COPY ./root/usr/bin /usr/bin
 # Copy the S2I scripts from the specific language image to $STI_SCRIPTS_PATH.
 COPY ./s2i/bin/ /usr/libexec/s2i
 
-# TODO: use .NET package (aspnetcore-runtime-6.0).
+# TODO: use .NET package (add: aspnetcore-runtime-6.0, remove:libicu).
+# Install packages:
+# - aspnetcore-runtime-*: provides the .NET runtime and ASP.NET Core shared framework.
+# - findutils: provides 'find' which is used by the 'assemble' script.
+# - tar: provides 'tar' which is needed by the Microsoft debugger install script (https://aka.ms/getvsdbgsh).
```
aslicerh, line 53: You mentioned `tar` here, but that has been removed from being installed below.

tmds: The comment was out-of-date. `tar` comes with the base image. The test verifies we can install `vsdbg`.

aslicerh (approved):
LGTM. Thanks.

PR #391, opened 2021-11-10
Add files for centos container pipeline integration to 6.0

aslicerh:


aslicerh:
After talking with @tmds and @omajid, it was decided to point centos users to using the ubi images instead, so closing this PR without merging.

PR #393, opened 2021-11-12
Update imagestreams and README for 6.0.

tmds:


tmds:
I verified the image streams work with OpenShift.

PR #394, opened 2021-11-12
Update imagestreams and README for 6.0.

tmds:
Add missing dotnet-runtime tags for 6.0.
Add 6.0 to the _centos streams. These are used by OKD.

@yselkowitz this fixes the issues you reported in https://github.com/redhat-developer/s2i-dotnetcore/pull/392#issuecomment-967081406.


tmds:
I've verified the imagestreams work with OpenShift.

Issue #406, opened 2022-06-08
npmrc config cannot be set


ascholz:
Hello,

I am currently trying to inject an nmprc config into the ubi8/dotnet-60 container at build time. I used the following guide to do this:

https://docs.openshift.com/container-platform/3.11/security/build_process.html#security-build-securing-inputs-during-builds

Here is the npmrc:

> registry=https://XXX
_auth=XXX
email=XXX
always-auth=true

Unfortunately it still tries to download packages from https://registry.npmjs.org. I also tried to set the NPM_MIRROR but that didn't help either.

Here is the log:

>Adding cluster TLS certificate authority to trust store
I0601 11:14:22.048972       1 builder.go:420] openshift-builder 4.6.0-202204261127.p0.g263041a.assembly.stream-263041a
[...]

PR #408, opened 2022-06-08
Move NPM flag checks in assembly file outside of DOTNET_NPM_TOOLS check.

aslicerh:
This should help resolve issue #406 by making NPM_MIRROR work as expected.

### DIFF
--- a/3.1/build/s2i/bin/assemble
+++ b/3.1/build/s2i/bin/assemble
@@ -115,21 +115,20 @@ if [ "$BUILD_TYPE" == "source" ]; then
   source /opt/app-root/etc/trust_ssl_dirs
 
   # npm
+  if [ ! -z $HTTP_PROXY ]; then
+    echo "---> Setting npm http proxy"
+    npm config set proxy $HTTP_PROXY
+  fi
+  if [ ! -z $HTTPS_PROXY ]; then
+    echo "---> Setting npm https proxy"
+    npm config set https-proxy $HTTPS_PROXY
+  fi
+  if [ ! -z $NPM_MIRROR ]; then
+    echo "---> Setting npm mirror"
+    npm config set registry $NPM_MIRROR
+  fi
   if [ -n "${DOTNET_NPM_TOOLS}" ]; then
     echo "---> Installing npm tools..."
-
-    if [ ! -z $HTTP_PROXY ]; then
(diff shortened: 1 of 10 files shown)


ascholz:
Question wouldn't it make sense to specify a login parameter as well? Because some registry need a login possibility.

https://stackoverflow.com/questions/23460980/how-to-set-npm-credentials-using-npm-login-without-reading-from-stdin

3.1/build/s2i/bin/assemble
```diff
@@ -115,22 +115,21 @@ if [ "$BUILD_TYPE" == "source" ]; then
   source /opt/app-root/etc/trust_ssl_dirs
 
   # npm
+  if [ ! -z $HTTP_PROXY ]; then
+    echo "---> Setting npm http proxy"
+    npm config set proxy $HTTP_PROXY
```
tmds, line 120: This will cause `npm` config steps to show up on .NET builds that don't use npm but have `HTTP(S)_PROXY` set.

Suggestion: make these steps conditional on `DOTNET_NPM_TOOLS` or `NPM_MIRROR` being set.

tmds:
> Question wouldn't it make sense to specify a login parameter as well? Because some registry need a login possibility.

Is this a use-case you have? If so, please create a separate issue for it.

tmds:
@aslicerh can you add some test? For example, something that check the s2i build output has some expected lines that related to these envvars.

ascholz:
> > Question wouldn't it make sense to specify a login parameter as well? Because some registry need a login possibility.
> 
> Is this a use-case you have? If so, please create a separate issue for it.

Done :-) [NPM config to set login parameter](https://github.com/redhat-developer/s2i-dotnetcore/issues/410)

3.1/build/s2i/bin/assemble
aslicerh: Ok, I just went ahead and moved the proxy checks back under NPM_TOOLS.

tmds: We can do the `npm config` stuff for HTTP_PROXY and HTTPS_PROXY if either DOTNET_NPM_TOOLS or NPM_MIRROR are set.

aslicerh:
Ok, I'm going to merge this in. If we need more adjustments, those can come later.

PR #421, opened 2022-10-04
7.0: remove scripts for RHEL7 software collections.

tmds:
We're no longer using these (since 6.0).


aslicerh:
Ah, yes, that's right. Looks good to me.

PR #423, opened 2022-10-05
imagestreams: add label that tells OpenShift images are provided by Red Hat.

tmds:


tmds:
Visually this will add a _Provided by Red Hat_ to the _Builder Image_ as you can see here for the _Helm Chart_.

image

omajid:
Is there some tool we can run to catch other metadata we should define? And/or some way to make sure we are defining known metadata in all our images?

aslicerh:
@omajid There's an auto-check for the minimum necessary labels required for RHEL images. I don't know of anything for other tool compatibility though (like for HELM charts).

PR #433, opened 2022-11-17
Add explicit license label instead of inheriting it fro the base image.

aslicerh:
This was requested by the RHEL container program.


omajid (approved):
LGTM!

PR #440, opened 2022-12-22
.NET 7: add missing timezone data.

tmds:
For .NET 7, we switched from the regular ubi base image to ubi-minimal. Unlike the ubi image, the minimal base does not include timezone data by default.

This brings back the timezones.


aslicerh (approved):
LGTM

PR #452, opened 2023-04-05
7.0: handle both cases of: reinstalling, as updating the timezone data.

tmds:
The command needed to install the timezone data
depends on the version of the package that is currently available in the base image.


tmds:
This is based on @omajid's question in https://github.com/redhat-developer/s2i-dotnetcore/pull/451#issuecomment-1497499539.

I don't know how to check this myself. Based on the answer on the bugzilla ticket, the command needed depends on the version installed in the base image.

Rather than changing the command back later, we can try both commands.

omajid (approved):
Thanks for digging into this and the followup!)

PR #453, opened 2023-04-19
Switch over to NodeJS 16, as NodeJS 14 is going EOL.

aslicerh:
This also fixes the issues with the angular and react templates, even on aarch64 and s390x.


tmds:
Thanks.

Do you get notified when these packages go EOL?

When is the EOL date of Node 16 compared to .NET 6?

aslicerh:
@tmds I don't, but a combination of a reminder I set up myself, and a user asking about this, triggered it.

.NET 6 goes EOL November 12th of 2024.

NodeJS v16 goes EOL September 11th of 2023.

NodeJS v18 goes EOL April 30th of 2025.

NodeJS v20 doesn't go LTS until October 24th, 2023, and goes EOL April 30th of 2026.

tmds:
If the templates work fine, let's move to Node 18, so we don't need to change it again for .NET 6.

omajid (approved):
Thanks!

PR #461, opened 2023-07-31
Bump the NodeJS version in the .NET 6 container to NodeJS v18.

aslicerh:
- Remove python3-systemd from the expected package list.
- Bump the expected versions to 6.0.20/6.0.120


omajid (approved):
LGTM! Thanks

Issue #464, opened 2023-08-16
trust_ssl_dirs, realpath and the service CA


cajosc:
We're running a .NET application which does HTTPS requests to other pods in the same Openshift cluster. These are secured using the service CA, and we add `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` to _DOTNET_SSL_DIRS_. The use of _realpath_ [here](https://github.com/redhat-developer/s2i-dotnetcore/blob/529c5c0c8317ee20f8a9c44488ce3d462bcf90cb/7.0/runtime/contrib/etc/trust_ssl_dirs#L44) (and [here](https://github.com/redhat-developer/s2i-dotnetcore/blob/529c5c0c8317ee20f8a9c44488ce3d462bcf90cb/7.0/runtime/contrib/etc/trust_ssl_dirs#L36)) causes the symlink in `/opt/app-root/ssl_dir` to target `service-ca.crt` in the timestamped directory `/var/run/secrets/kubernetes.io/serviceaccount/..YYYY-MM_DD_HH_mm_ss.NNNNNNNN` instead.

Since `/var/run/secrets/kubernetes.io/serviceaccount` is a projected volume with, among other things, a _serviceAccountToken_ source with an expiration time of 3607 seconds, the symlink stops working after that time when the timestamped directory is replaced.

```yaml
spec:
  containers:
  - name: dotnet-app
    volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: kube-api-access-chtld
      readOnly: true
...
  volumes:
  - name: kube-api-access-chtld
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          expirationSeconds: 3607
          path: token
      - configMap:
          items:
          - key: ca.crt
            path: ca.crt
          name: kube-root-ca.crt
      - downwardAPI:
          items:
          - fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
            path: namespace
      - configMap:
          items:
          - key: service-ca.crt
            path: service-ca.crt
          name: openshift-service-ca.crt
```

Is it really necessary to use _realpath_ to resolve symlinks?

aslicerh:
Ok, so there are a few different things here and let me make sure I understand what is going on and what is expected.

You have a file, `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt`, that is added to `DOTNET_SSL_DIRS`.
[...]

PR #466, opened 2023-08-21
Add --no-symlinks to calls to realpath when linking SSL certs.

aslicerh:
This is in reference to issue #464 .


tmds:
> This is in reference to issue https://github.com/redhat-developer/s2i-dotnetcore/issues/464 .

@aslicerh GitHub can infer the relation with the issue if you use one of these keywords:

Fixes/Closes/Resolves https://github.com/redhat-developer/s2i-dotnetcore/issues/464.

Then it will close the issue when the PR is merged.

Can you make the same change to 6.0?

tmds (approved):

tmds:
@aslicerh thanks for implementing this! If the existing tests pass, go ahead and merge.

Issue #455, opened 2023-05-15
[.NET 8] remove HTTP_PROXY to http_proxy assignment.


tmds:
This is no longer necessary since the libcurl based handler was removed in .NET 5:

https://github.com/redhat-developer/s2i-dotnetcore/blob/49993a6f6061f60c70612224446df0d92e1b6407/7.0/runtime/root/usr/bin/container-entrypoint#L7-L13

This issue tracks removing it for .NET 8 images.

cc @omajid @aslicerh

PR #471, opened 2023-10-03
8.0: remove unneeded mapping from HTTP_PROXY to http_proxy.

tmds:
This mapping was needed when .NET's HTTP stack used curl under the hood. The curl handler was removed in .NET 5, so we no longer need this.

Fixes https://github.com/redhat-developer/s2i-dotnetcore/issues/455.


tmds:
Thanks for taking a look @aslicerh!

Issue #448, opened 2023-03-24
[.NET 8] align with some Microsoft changes


tmds:
As we want to facilitate users switching between Red Hat and Microsoft .NET images, we may want to adopt some changes Microsoft is making, after assessing the backwards compatibility impact.

From https://devblogs.microsoft.com/dotnet/securing-containers-with-rootless/:

- Rename the default user to `app`
- Use `ASPNETCORE_HTTP_PORTS` instead of `ASPNETCORE_URLS`
- Add `APP_UID`

cc @omajid @aslicerh

PR #472, opened 2023-10-04
Use ASPNETCORE_HTTP_PORTS and set APP_UID.

tmds:
This aligns with changes made to Microsoft images for .NET 8.

Fixes https://github.com/redhat-developer/s2i-dotnetcore/issues/448.


8.0/runtime/test/aspnet-hello-world/Program.cs
```diff
@@ -1,18 +1,6 @@
-﻿using Microsoft.AspNetCore.Hosting;
-using Microsoft.AspNetCore.Builder;
-using Microsoft.AspNetCore.Http;
+﻿var builder = WebApplication.CreateBuilder(args);
```
tmds, line 1: The ASPNETCORE_HTTP_PORTS was introduced in a backwards compatible way.

It seems the backwards compatibility does not include the old WebHost API which this example was still using.
I've created an issue: https://github.com/dotnet/aspnetcore/issues/51125.
We should follow what comes out.

tmds: I don't like introducing a non-breaking change when it is not needed,
though I prefer to follow what Microsoft does here, and document the breaking change.

If Microsoft makes some further changes to avoid the breaking change, we can sync again for those.

tmds:
@aslicerh @omajid can you take a look and approve the PR?

aslicerh (approved):

tmds:
@aslicerh, thanks!

PR #474, opened 2023-10-11
Revert adopting ASPNETCORE_HTTP_PORTS.

tmds:
Microsoft .NET 8 images adopt ASPNETCORE_HTTP_PORTS in favor of ASPNETCORE_URLS.

This new envvar provides a slightly user-friendlier alternative to control the port ASP.NET Core applications listen on.

For Microsoft images this change goes together with changing from port 80 to port 8080.
This means users are anyhow confronted with a breaking change.

For Red Hat images we were already using port 8080 and continue using that same port.

When updating our images to use ASPNETCORE_HTTP_PORTS instead of ASPNETCORE_URLS we learned that the new environment variable is not recognized by ASP.NET Core applications that use the older WebHost API. This causes such applications to bind to port 5000 instead of port 8080.

To avoid breaking these applications, we revert the changes to adopt ASPNETCORE_HTTP_PORTS and stick to using
ASPNETCORE_URLS.

Partly reverts https://github.com/redhat-developer/s2i-dotnetcore/pull/472.


aslicerh (approved):

tmds:
Thanks for reviewing @aslicerh!

PR #500, opened 2024-12-17
Remove test_rpm_packages test.

tmds:
This removes the test_rpm_packages test which captures any changes in packages included in the .NET images. This test is giving some maintenance challenges and false positives.

Instead of this, we try to capture functionally in tests what we expect the image to provide, and rely on due diligence of reviewers when making changes to Containerfiles as how they affect backwards compatibility.


omajid (approved):
Thanks!

PR #512, opened 2025-08-18
Sync with downstream.

tmds:


tmds:
Thanks @omajid!

Let's see how some of the fix-ups now flow downstream...

phracek:
I will look how upstream -> downstream works and if MR was created or updated.

PR #515, opened 2025-09-19
Update container labels for .NET 10.

tmds:
We no longer set the 'version' label so that the label from the base image is preserved.

The 'dotnet_version' label is used for the version of the component.
For the SDK image, it is the full SDK version, like 10.0.100.
For the runtime image, it is the full ASP.NET Core runtime version, like 10.0.0.

The 'dotnet_version_major_minor' contains only the major.minor part, like 10.0.

### DIFF
--- a/10.0/runtime/Dockerfile.rhel9
+++ b/10.0/runtime/Dockerfile.rhel9
@@ -39,8 +39,8 @@ LABEL name="ubi9/dotnet-100-runtime" \
       com.redhat.component="dotnet-100-runtime-container" \
       com.redhat.license_terms="https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI" \
       dotnet_version_major_minor="10.0" \
-      # .NET version for this component
-      dotnet_version="$ASPNET_VERSION"
+      dotnet_version="$DOTNET_VERSION" \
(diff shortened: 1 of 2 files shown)


phracek (changes requested):
LGTM.

10.0/runtime/Dockerfile.rhel9
```diff
@@ -38,8 +38,9 @@ LABEL name="ubi9/dotnet-100-runtime" \
       summary=".NET 10 runtime" \
       com.redhat.component="dotnet-100-runtime-container" \
       com.redhat.license_terms="https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI" \
-      version="10.0" \
-      dotnet_version="$DOTNET_VERSION"
+      dotnet_version_major_minor="10.0" \
+      # .NET version for this component
+      dotnet_version="$ASPNET_VERSION"
```
phracek, line 43: Is this really $ASPNET_VERSION? It was $DOTNET_VERSION previously. Onnly check

tmds: I changed the labels so the envvars and labels have the same names and values. That avoids any ambiguity.

PR #523, opened 2025-10-24
10.0: review image labels.

tmds:


tmds:
@omajid this is based on my current understanding of the labels. If we get additional insights, we can update this in the next week(s). ptal.

omajid (approved):

PR #524, opened 2025-11-03
10.0: revert changes to com.redhat.component.

tmds:
The Red Hat build pipeline requires derived images to have different values from their base image.


tmds:
@omajid fyi, this reverts changes to com.redhat.component label that were made in https://github.com/redhat-developer/s2i-dotnetcore/pull/523.

PR #534, opened 2026-05-07
test: replace myget.org feed with feedz.io

tmds:
Replace the myget.org NuGet feed with feedz.io due to frequent internal server error responses from myget.org.


omajid:
[test]

omajid (approved):
Thanks!

PR #535, opened 2026-06-30
Initial .NET 11.0 changes.

tmds:
The usual changes for a new .NET version.

Tests pass with a preview5 build.


11.0/aspnet/test/aspnet-hello-world/app.tar.gz
omajid, line 1: Is there something we can/should do to assert/verify that this is not malicous and generated by a known by a known set of commands?

omajid (approved):
Looks good to me. I wish we had a good way to verify/confirm the binaries.

tmds:
> I wish we had a good way to verify/confirm the binaries.

Such things come down to what you trust. You could download the tar.gz files, decompile the IL and see if it doesn't have anything suspicious. I don't think it's worth automating such things because we don't have to do this often and it requires extending your trust to that automation.

Thanks for the review!

